PR #2, opened 2018-09-06
[Do Not Merge] Migrate repo

wbuchwalter:
Do not merge yet, unless we are OK with having an MIT licenced project under Kubeflow org.
I will try to get the licence changed to Apache 2 ASAP.

Instructions to try it out: https://github.com/wbuchwalter/fairing/tree/master/examples/kubeflow-jupyter-notebook

Then in a notebook cell, simple do:
`
!pip install fairing
`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/2)

PR #5, opened 2018-11-02
Migrate wbuchwalter/fairing

wbuchwalter:
Follow up to https://github.com/kubeflow/fairing/pull/2 with correct licence change applied.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/5)


wbuchwalter:
@jlewi I will let you merge this one once you have review that licensing is correct.
Note that in this state, fairing does not work with the current version of Kubeflow because of JupyterHub described in https://github.com/kubeflow/fairing/issues/6.

I am working on a fix. I will open a PR here once it is merged.

Also not that @ashahba has been working on a PR to add a new builder using configmaps instead of Knative. This PR will be transferred here as well so it can be discussed by the community.

jlewi:
/assign @r2d4 

@r2d4 Can you take a lead on reviewing this please?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/will/envs/DoWhy-zTd0ka9x/bin/python"
```
r2d4: I think this file can be removed

Dockerfile.dev
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /opt/fairing
+
+# First copy setup.py and do a pip install -e to 
+# only install dependencies. This will make 
+# subsequent docker builds much faster
+COPY setup.py /opt/fairing/
+RUN pip install -e .
+
+COPY ./ /opt/fairing
+RUN python ./setup.py install
```
r2d4: nit: add a newline

Dockerfile.jupyter
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /opt/fairing
+
+# First copy setup.py and do a pip install -e to 
+# only install dependencies. This will make 
+# subsequent docker builds much faster
+COPY setup.py /opt/fairing/
+RUN pip install -e .
+
+COPY ./ /opt/fairing
+RUN python ./setup.py install
```
r2d4: nit: add a newline

Makefile
```diff
@@ -0,0 +1,22 @@
+.PHONY: dev push_image
+
+dev:
+	@if [ -z "${FAIRING_DEV_DOCKER_USERNAME}" ]; then\
+		echo "FAIRING_DEV_DOCKER_USERNAME is unset. Set it to your docker username to use this command"; \
+	else \
+		pip install . ;\
+		docker build -t ${FAIRING_DEV_DOCKER_USERNAME}/fairing:latest -f Dockerfile.dev . ; \
```
r2d4: I have some ideas on how we could get rid of scripts like this, but for now you might want to avoid pushing anything with "latest" and tag with git commit sha.

Dockerfile.jupyter
```diff
@@ -0,0 +1,11 @@
+FROM tensorflow/tensorflow:latest-py3
```
r2d4: Let pin to a specific version of tf?

README.md
```diff
@@ -1,2 +1,230 @@
...
+    architecture=DistributedTraining(ps_count=2, worker_count=5),
+)
+class MyModel(object):
+    ...
+```
+
+Specify the number of desired parameter servers with `ps_count` and the number of workers with `worker_count`.
+Another instance of type master will always be created.
+
+See [https://github.com/Azure/kubeflow-labs/tree/master/7-distributed-tensorflow#modifying-your-model-to-use-tfjobs-tf_config](https://github.com/Azure/kubeflow-labs/tree/master/7-distributed-tensorflow#modifying-your-model-to-use-tfjobs-tf_config) to understand how you need to modify your model to support distributed training with Kubeflow.
```
r2d4: Maybe this README should be moved somewhere within kubeflow/*?

README.md
```diff
@@ -1,2 +1,230 @@
...
+Your model needs to be defined as a class to work with `fairing`. 
+
+This limitation is needed in order to enable usage of more complex training strategies and simplify usage from within a Jupyter Notebook.
+
+Following are a series of example that should help you understand how fairing works.
+<!-- The train decorator 
+* `package`: Defines the repository (this could be your DockerHub username, or something like `somerepo.acr.io` on Azure for example) and name that should be used to build the image. You can control wether you want to publish the image by setting `publish` to `True`.
+* `strategy`: Specify which training strategy should be used (more details below).
+* `architecture`: Specify which architecture should be used. (more details below)
+* `tensorboard`: [Optional] If specified, will spawn an instance of TensorBoard to monitor your trainings
```
r2d4: Should this be the responsibility of `fairing`? (Not a leading question, I'm still understanding the whole project).

README.md
```diff
@@ -1,2 +1,230 @@
...
+class MyModel(object):
+    def train(self):
+       # training logic
+```
+
+
+### Distributed Training
+
+Using Kubeflow, we can also ask `fairing` to start [distributed trainings](https://www.tensorflow.org/deploy/distributed) instead.
+Simply import `DistributedTraining` architecture insteda of the `BasicArchitecture`:
```
r2d4: Shouldn't distributed training be the default? Why would you not want to run it distributed if you're running it on kubeflow?

README.md
```diff
@@ -1,2 +1,230 @@
...
+
+Complete example: [examples/hyperparameter-tuning/main.py](./examples/hyperparameter-tuning/main.py)
+
+#### Population Based Training
+
+We can also ask `fairing` to train our code using [Population Based Training](https://deepmind.com/blog/population-based-training-neural-networks/).
+
+This is a more advanced training strategies that needs hook into different lifecycle steps of your model, thus we need to define several additional method into our model class.
+
+A multiple read/write PVC name needs to be pass to the `PopulationBasedTraining` strategie. This is used to store and exchange the different models generated by our training to enable the `explore/exploit` mechanism of Population Based Training.
```
r2d4: Can you explain why the PVC is needed?

r2d4: I was thinking that fairing would be responsible for converting the notebook to a container, and then spawning the correct TF/Pytorch job through the TFJob CR.

examples/distributed-training/main.py
```diff
@@ -0,0 +1,248 @@
...
+                        summary, _ = sess.run(
+                            [self.merged, self.train_step], feed_dict=feed_dict(True))
+                        train_writer.add_summary(summary, i)
+            train_writer.close()
+            test_writer.close()
+
+
+if __name__ == '__main__':
+    model = MyModel()
+    model()
```
r2d4: nit: newline

examples/distributed-training/requirements.txt
```diff
@@ -0,0 +1 @@
+tensorflow
```
r2d4: pin this to a version

examples/hyperparameter-tuning/requirements.txt
```diff
@@ -0,0 +1 @@
+tensorflow
```
r2d4: pin this to a version

examples/kubeflow-jupyter-notebook/README.md
```diff
@@ -0,0 +1,66 @@
...
+### Kubeflow
+
+Kubeflow needs to be pre-installed in the cluster.
+
+### Knative Build Component
+A Knative Build controller needs to be deployed in the cluster beforehand to handle in-cluster builds.  
+See: [Installing Knative Build Component](https://github.com/knative/docs/blob/master/build/installing-build-component.md)
+
+
+### Container Registry Secret
```
r2d4: I think that we should reuse the user service account that gets created instead

examples/kubeflow-jupyter-notebook/README.md
```diff
@@ -0,0 +1,66 @@
+# Using `Fairing` with Kubeflow's TfJob inside a Jupyter Notebook
+
+## Requirements
+
+### Kubeflow
+
+Kubeflow needs to be pre-installed in the cluster.
+
+### Knative Build Component
```
r2d4: I think that Knative is a heavy dependency to build images on the cluster. If we're just appending python files, we can do it much simpler and quicker without knative.

fairing/architectures/kubeflow/distributed.py
```diff
@@ -0,0 +1,49 @@
+from fairing.architectures.kubeflow.basic import BasicArchitecture
+from fairing.backend.kubeflow import KubeflowBackend
+
+
+class DistributedTraining(BasicArchitecture):
+    def __init__(self, ps_count, worker_count):
+        self.ps_count = ps_count
+        self.worker_count = worker_count
+
+    def add_jobs(self, svc, count, img, name, volumes, volume_mounts):
```
r2d4: WDYT about using the k8s clients directly here?

fairing/backend/native.py
```diff
@@ -0,0 +1,76 @@
...
+        ]
+
+        svc["serve"] = {
+            "name": "{}-fairing-serving".format(name),
+            "public": True
+        }
+        return svc
+    
+    def stream_logs(self, image_name, image_tag):
+        mp_client = MetaparticleClient()
```
r2d4: Why the dependency on metaparticle? It seems a bit unnecessary

fairing/builders/knative/knative.py
```diff
@@ -0,0 +1,103 @@
...
+        params = [BuildTemplateSpecParameter(name='IMAGE', description='The name of the image to push'),
+                  BuildTemplateSpecParameter(name='TAG', 
+                                             description='The tag of the image to push',
+                                             default='latest'),
+                  BuildTemplateSpecParameter(name='DOCKERFILE',
+                                             description='Path to the Dockerfile to build',
+                                             default='/src/Dockerfile')]
+        volume_mount = kubernetes.client.V1VolumeMount(name='src', mount_path='/src')
+        steps = [BuildTemplateSpecStep(name='build-and-push',
+                                       image='gcr.io/kaniko-project/executor',
```
r2d4: nit: pin to a version.

If we're hardcoding kaniko here, it might be simple to just launch the kaniko job directly instead of calling it through knative. But it might be easier to skip kaniko and build CRD altogether.

Makefile
wbuchwalter: This specific script is to make easier to help developers try their changes without having to go through the pip release process.
You would set `FAIRING_DEV_DOCKER_USERNAME` to your username and this would push a dev version to your dockerhub. This image is then pulled instead of the regular base image.
So latest is needed so that you don't have to go modify this script on every change.

examples/kubeflow-jupyter-notebook/README.md
wbuchwalter: Knative is the solution that came out after discussing with @jlewi.
You can see some of the context for that in this thread: https://github.com/kubeflow/kubeflow/issues/1240#issuecomment-408878266

Note that @ashahba has a PR opened on the old repo to use configmaps directly and bypass the build step entirely (https://github.com/wbuchwalter/fairing/pull/27). I ll move this PR here once this one get merged. However this approach does not allow for a lot of flexibility since you cannot customize the remote environment.

Curious to hear your idea.

fairing/backend/native.py
wbuchwalter: Context for metaparticle here: https://github.com/kubeflow/kubeflow/issues/1240#issuecomment-408493465.
However I am not convinced anymore that this is the right approach since metaparticle is not maintained it seems.
We can remove it in a later PR.

examples/kubeflow-jupyter-notebook/README.md
wbuchwalter: I'm not sure I understand with service account you are referring to?

wbuchwalter:
@r2d4 Addressed some of your comments.
For everything related to removing metaparticle and using k8s client directly instead, I agree with the direction however I would like that to be done in a follow-up PR so that I don't block all the progress of other PRs.
Same applies to Knative, I am not specifically attached to Knative and I am curious to hear what you propose, but it would be nicer to discuss this in a separate issue and implement what is decided in a follow-up PR.

What do you think?

r2d4:
Yep SGTM.

/approve
/lgtm

PR #15, opened 2018-11-08
Adding a configMap builder and architecture

ashahab:
This adds a configMap builder and architechture to allow using
 configmaps for the code for notebooks during distributed
 run. This avoids having to build images and therefore can aid
 distributed training during early exploration.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/15)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 .metaparticle
 examples/*/Dockerfile
-
+.idea
 .vscode
 
 # Byte-compiled / optimized / DLL files
--- a/fairing/architectures/kubeflow/cm.py
+++ b/fairing/architectures/kubeflow/cm.py
@@ -1,0 +1,68 @@
+from fairing.architectures.kubeflow.basic import BasicArchitecture
+from fairing.backend.kubeflow import KubeflowBackend
+from fairing.notebook_helper import get_notebook_name
+from fairing.kubeclient import KubeClient
+from jinja2 import Template
+import subprocess
+import os
+import logging
+import yaml
+import json
+logger = logging.getLogger(__name__)
+
+class CMTraining(BasicArchitecture):
+    def __init__(self, ps_count, worker_count, notebook):
+        self.ps_count = ps_count
+        self.worker_count = worker_count
+        self.notebook = notebook
+        self.client = KubeClient()
+
+    def add_jobs(self, svc, count, repository, img, name, volumes, volume_mounts):
+        worker_image = os.environ['WORKER_IMAGE_NAME'] or img
+        ps_image = os.environ["PS_IMAGE_NAME"]
+        nb_name = self.notebook or get_notebook_name()
+        nb_full_path = os.path.join(os.getcwd(), nb_name)
+        cmd = "jupyter nbconvert --to python {} --output /tmp/code".format(nb_full_path).split()
+        subprocess.check_call(cmd)
+
+        with open('/tmp/code.py', 'rb') as f:
+            code = f.read()
+        configMap = {
+            "apiVersion": "v1",
+            "kind": "ConfigMap",
+            "metadata": {
+               "name": name
+            },
+            "data": {
+                "code.py": code
+            }
+        }
+        svc["configMap"] = configMap
+        volume_mounts = volume_mounts or []
+        volume_mounts.append({
+            "name": "code",
+            "mountPath": "/code"
+        })
+        volumes = volumes or []
+        volumes.append({
+            "name": 'code',
+            "configMap": {
+                "name": name
+            }
+        })
+        cmd = "python /code/code.py".split()
+        # TODO: This should come from configs
+        # append configmap to volume and volumeMounts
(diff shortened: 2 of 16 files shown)


wbuchwalter:
I had a chat with @r2d4 yesterday, one concern was around etcd.

Etcd does not allow objects to be more than 2MB, which for regular notebooks or python files is enough, but what happens if I have a large JSON or pickle file as part of my code? 

Also, can you expand a bit on your motivations for this builder ? (i.e. what pain points is this solving compared to Knative builder). 
One of @r2d4 main focus will be refactoring and improving builders so this would be pretty valuable feedback.

ashahab:
@wbuchwalter @r2d4 
I understand the concern about etcd, and it is entirely fine for us to say "this only works with files smaller than 2MB, otherwise this would require knative, or some persistent volume support".
My motivations are: 
* Ensuring running jobs from notebook is easy for ML engineers who may not be experts at container images(that's the situation with a majority of ML engineers at Linkedin). 
* Reduce several points of failure (knative issue, or issue with the tooling that builds the images, or having to manage a separate security model to securely build the images). For example our RHEL7 os cannot handle user namespaces very well and so still has issues with a builder like img. At the kubeflow contributor meet, several people expressed reservations about knative, and so I want to dig into that more before committing to it.
* Allows a Kubernetes-novice to quickly start without relying on setting up knative
* More intuitive than knative, since this is essentially running "the notebook image you used" + "the notebook code you ran"
* I think it goes with the spirit of fairing not being opinionated about whether to use Kubernetes or not. Here we are expanding that openness of opinion to whether an image builder is needed or not. The CM builder can work fine with just a bare-bones CRI-o support on the nodes.

r2d4:
@ashahab would the code I just merged satisfy your use case? My next step is integrating it as a builder in the fairing python library.

The method behind that is a totally unprivileged (in all senses) and quick builder (builds + push in ~3-4 seconds) with the limitation that there is no RUN command.

For a little more info, it takes a tarball (which contains the converted notebook python file) and appends it to the container as a new image. It can be overlaid over any base image without pulling the image (which makes it just as quick for a 1MB base image as 1GB base image) and pushed with only pushing the changed layer (the python file).

r2d4:
Yes, there are no requirements for this builder (except that you must have the right credentials somewhere to push)

ashahab:
Hi Matt,
I saw your PR:
https://github.com/kubeflow/fairing/commit/668f5e90daf2f84d5ff3f76c28fdf3fffbc32deb
Let me know if that's the correct one to look at.
I like the fairing go builder and am eager to try it out.
However, setting up image pushing authn integration is fairly complicated
from inside a container.
We are also running rootless notebook with only a few users(uid1000 only
actually) defined for our notebooks, so I want our users to have a taste of
fairing before we have to worry about whether our notebook container have
enough privileges to run mutate, or if we are capable of pushing images in
a scalable way.

The advantage of the Configmap Builder is, if it breaks, one of the key
capabilities of the Kubernetes cluster is broken: either etcd has stopped
supporting creation of configmap, or the current jupyter notebook image
cannot be used to spawn containers. Nothing else is necessary. That's why I
feel it's needed as a gentle introduction to fairing. I find fairing to be
very useful in our ML pipelines and it would not be good if there's a large
investment in images needed for users to take advantage of it.
Abin

ashahab:
I will say that in future I'd love to create/try out a
persistent-volume-based solution to get around the 2MB problem.
Configmap was my first attempt in that direction because even setting up
and operating a PV solution across 1000s of nodes can be daunting for some
eager users of fairing.
Abin

r2d4:
@ashahab ping this needs a rebase :+1:

ashahab:
Yup, I'm on vacation for Thanksgiving, will fix it when I'm back.

jlewi:
@ashahab It would be great to get LinkedIn to try this and Kubeflow out. 

> Configmap was my first attempt in that direction because even setting up
and operating a PV solution across 1000s of nodes can be daunting for some
eager users of fairing.

Would users have to setup a PV solution? Does LinkedIn have a shared filesystem e.g. NFS that could be mounted via PV? Or an object store or HDFS?

@wbuchwalter  Using a ConfigMap for code seems reasonable since I expect code to be under 1 Mb. For json objects and other data I don't think ConfigMap is a good solution. PV is a better option. In the case of S3/GCS/HDFS we could just copy the data to/from the filestore. We could potentially use init containers or side cars to make that pattern easier.

Another option is what @vkoukis @ioandr and the folks at Arikto are building. That makes it easy to snapshot volumes and replicate them to pods using a CSI.

wbuchwalter:
@jlewi Yep. I am almost done with a pretty significant refactoring (https://github.com/kubeflow/fairing/pull/20#discussion_r235133570) that should make it much easier to integrate new builders, including this one.

wbuchwalter:
@ashahab As discussed on Slack, I have reimplement this on top of the new APIs in #21. 
Let's continue the discussion there.

PR #17, opened 2018-11-12
Update readme & url in setup.py

wbuchwalter:
* Fixes the url in `setup.py`
* Removes mention of serving from the doc for now. We may want to add that back later when/if we implement a more solid solution but for now it's a bit misleading.

cc @r2d4

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/17)


r2d4:
/lgtm
approve

PR #18, opened 2018-11-12
Add Python2 support

wbuchwalter:
This should enable using fairing from python 2.7 and 3.3+

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/18)


wbuchwalter:
@r2d4 this should be ready for review.

PR #20, opened 2018-11-19
Refactor APIs

### LINKED ISSUE #8: Remove Metaparticle
Metaparticle isn't maintained anymore, so we probably should just used the kubernetes client directly instead.

wbuchwalter:
This PR is a pretty big refactor.

### Main changes
* The concept of `architecture` and `strategy` have been removed. This might lead to slightly more code duplication, but a much more flexible code base and much easier to understand.
* Introduce the concept of  `Deployment` and `Runtime`. `Deployment` describe the behavior while packaging and deploying the code while `Runtime` describe the behavior on the remote target (i.e. a TFJob on Kubernetes). Currently `Runtime` is pretty simple, but will be useful once we re-introduce hyperparameter tuning and Population based training
* Each provider (native, kubeflow) now can export several decorators (e.g. `Training` and `DistributedTraining`) instead of always working with `Train`. Issue with the previous approach is that `Train` ended up having tons of parameters which may or may not apply to your particular training, it was confusing.
* Remove dependency on metaparticle (fixes #8) and allow each step of the build/deployment to easily modify the generated k8s objects. This should make https://github.com/kubeflow/fairing/pull/15 much easier to achieve.
* Introduce a `Config` object which can be used to specify which `builder` to use. With this Builders can now take whatever argument they need without having to surface them via the training decorator.

### Temporary
 This features where removed because the current approach was not right, but these are definitly valuable features to have, I will open an issue for each of them to discuss how to reintroduce them properly:
* Remove hyperparameter tuning
* Remove TensorBoard
* Remove Population Based Training (Not sure if we want to reintroduce this for now)

cc @r2d4 @ashahab

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/20)


fairing/training/native/__init__.py
```diff
@@ -0,0 +1,10 @@
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
+from future import standard_library
+standard_library.install_aliases()
+
+from .deployment import NativeDeployment
+from .runtime import BasicNativeRuntime
+from .decorators import Training
```
r2d4: nit newline

fairing/training/native/decorators.py
```diff
@@ -0,0 +1,57 @@
...
+    def _train(self, user_object):
+        runtime = BasicNativeRuntime()
+        runtime.execute(user_object)
+
+    def _deploy(self, user_object):
+        deployment = NativeDeployment(self.namespace, self.runs)
+        deployment.execute()
+
+
+# class HPTuning(base.TrainingDecoratorInterface):
```
r2d4: is this meant to be commented out?

fairing/training/native/deployment.py
```diff
@@ -0,0 +1,100 @@
...
+from fairing import config
+from fairing import utils
+from fairing.training import base
+from fairing.backend import kubernetes
+
+logger = logging.getLogger(__name__)
+DEFAULT_JOB_NAME = 'fairing-job'
+
+class NativeDeployment(object):
+    """Handle all the template building for metaparticle api and calling mpclient
```
r2d4: Does this comment need to be updated?

fairing/training/native/deployment.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    def __init__(self, namespace, runs):
+        if namespace is None:
+            self.namespace = utils.get_default_target_namespace()
+        else:
+            self.namespace = namespace
+        
+        # Used as pod and job name
+        self.name = "{}-{}".format(DEFAULT_JOB_NAME, utils.get_unique_tag())
```
r2d4: v1.objectmeta has a nice field called "generateName" that will do this for you. That might be slightly simpler.

fairing/training/native/deployment.py
```diff
@@ -0,0 +1,100 @@
...
+            spec=job_spec
+        )
+
+    def get_logs(self):
+        self.backend.log(self.name, self.namespace)
+
+    def deploy(self):
+        """Handles communication with kubeclient to deploy 
+            the component needed for the training"""
+        self.backend.create_job(self.namespace, self.job_spec)
```
r2d4: nit: newline

fairing/training/native/runtime.py
```diff
@@ -0,0 +1,17 @@
...
+class BasicNativeRuntime(object):
+    """BasicNativeRuntime represents the behavior of the code while training
+        during a simple training job"""
+
+    def execute(self, user_object):
+        if 'build' in dir(user_object) and callable(getattr(user_object, 'build')):
+            user_object.build()
+        user_object.train()
+
+#class HPNativeRuntime
```
r2d4: remove comment?

fairing/builders/docker_builder.py
```diff
@@ -13,53 +13,80 @@
...
-            self.publish(full_image_name)
+      
+    def generate_pod_spec(self):
+        """return a V1PodSpec initialized with the proper container"""
 
-    def build(self, img, path='.'):
-        logger.warn('Building docker image {}...'.format(img))
-        if self.docker_client is None:
-            self.docker_client = APIClient(version='auto')
+        return client.V1PodSpec(
```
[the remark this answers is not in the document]
wbuchwalter: I set a `fairing-job-id` label on the `PodTemplateSpec` here: https://github.com/kubeflow/fairing/pull/20/files#diff-df5d1c0a2b7131d60669df7a4fc96720R70

fairing/training/native/decorators.py
wbuchwalter: Yes, I would like to open a discussion of how to imlement hyperparameter tuning properly before reintroducing it. There are a bunch of things that weren't right with the previous approach.
I can also remove the code entirely instead of commenting it if it's an issue.

wbuchwalter:
@r2d4 Thanks for the review!
I applied or responded to all your comments except for `generateName` as I don't have enough time to test it right away

r2d4:
Sounds good
/lgtm
/approve

r2d4:
Feel free to remove the /hold and merge when you want to

examples/knative-builder/main.py
```diff
@@ -41,7 +46,7 @@
                        'tensorflow/mnist/logs/fully_connected_feed/', os.getenv('HOSTNAME', ''))
 MODEL_DIR = os.path.join(LOG_DIR, 'model.ckpt')
 
-@Train(repository='<your-repository>', builder='knative')
+@native.Training()
```
ashahab: Is train() method in the decorated class required or optional now? Is there a way to demonstrate that through examples?

fairing/training/kubeflow/deployment.py
```diff
@@ -0,0 +1,52 @@
...
+class KubeflowDeployment(deployment.NativeDeployment):
+
+    def __init__(self, namespace, runs, distribution):
+        super(KubeflowDeployment, self).__init__(namespace, runs)
+        self.distribution = distribution
+
+    def deploy(self):
+        self.backend.create_tf_job(self.namespace, self.job_spec)
+    
+    def generate_job(self, pod_template_spec):
```
ashahab: There should be a way to generate a job config from a configmap, but it can be a different deployment class to be added later.  Ac configmap job spec allows users/admins to customize the spec.

examples/knative-builder/main.py
wbuchwalter: Currently it is still required. We can open a follow-up issue to discuss how to remove this requirement.

fairing/training/kubeflow/deployment.py
wbuchwalter: Yes I plan on opening a follow-up issue regarding admin config. I am not sure yet what are all the things an admin would want to configure that I think we need to discuss it a bit before implementation.

ashahab:
Some example things are other things in the pod spec(env, initContainers,
mounts, securityContext). Configmap allows us to not limit fairing
TFjobs to what we support today.

fairing/training/native/deployment.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def __init__(self, namespace, runs):
+        if namespace is None:
+            self.namespace = utils.get_default_target_namespace()
+        else:
+            self.namespace = namespace
+        
+        # Used as pod and job name
+        self.name = "{}-{}".format(DEFAULT_JOB_NAME, utils.get_unique_tag())
+        k8s_client.V1ObjectMeta.generate_name()
```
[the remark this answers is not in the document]
wbuchwalter: Nop, good catch.

wbuchwalter:
@r2d4 fixed. Can you re -apply lgtm when you have time?

PR #21, opened 2018-11-22
[WIP] Add ConfigMap builder

wbuchwalter:
Re-implementation of #15 following the refactoring in #20.
cc @ashahab 

In this current state, there is now way to specify different base images for worker and ps.
I will address that in a separate PR, as this need to be solved for all builders not just this one.

This PR also contains some unrelated changes such as renaiming `native` in `kubernetes`.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/21)

### DIFF
--- a/fairing/builders/config_maps.py
+++ b/fairing/builders/config_maps.py
@@ -12,19 +12,16 @@ from fairing.backend import kubernetes
 
 class ConfigMapBuilder(BuilderInterface):
 
-    def __init__(self, base_image, notebook_name=None):
+    def __init__(self, notebook_name=None):
         if notebook_name is None:
             self.notebook_name = notebook_helper.get_notebook_name()
         else:
             self.notebook_name = notebook_name
-        self.base_image = base_image
 
-    def execute(self, namespace, job_id):
+    def execute(self, namespace, job_id, base_image):
         nb_full_path = os.path.join(os.getcwd(), self.notebook_name)
-
         temp_dir = tempfile.mkdtemp()
         code_path = os.path.join(temp_dir, 'code.py')
-
         try:
             cmd = "jupyter nbconvert --to python {nb_path} --output {output}"
             .format(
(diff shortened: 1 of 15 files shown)


fairing/builders/config_maps.py
```diff
@@ -0,0 +1,61 @@
...
+from fairing import notebook_helper
+from .builder import BuilderInterface
+from fairing.backend import kubernetes
+
+
+class ConfigMapBuilder(BuilderInterface):
+
+    def __init__(self, notebook_name=None):
+        if notebook_name is None:
+            self.notebook_name = notebook_helper.get_notebook_name()
```
ashahab: Does this work? If not, we should make notebook_name required.

fairing/builders/config_maps.py
```diff
@@ -0,0 +1,61 @@
...
+            subprocess.check_call(cmd)
+            with open(code_path, 'rb') as f:
+                code = f.read()
+        finally:
+            shutil.rmtree(temp_dir, ignore_errors=True)
+
+        client.V1ConfigMap(
+            api_version="v1",
+            data={"code.py": code},
+            metadata=client.V1ObjectMeta(name=self.job_id)
```
ashahab: Is job_id a parameter or member? Who sets it?

fairing/builders/config_maps.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        k8s = kubernetes.KubeManager()
+        k8s.create_config_map(namespace, config_map)
+        return self._generate_pod_spec(job_id, base_image)
+
+    def _generate_pod_spec(self, job_id, base_image):
```
ashahab: Not sure if it would be this PR but would be great to provide this pod-spec using a configmap(that way specifying images, environments, and other mounts is completely upto the user).

fairing/builders/config_maps.py
wbuchwalter: It works in standalone notebook, not with JupyterHub.
I can improve the error handling in the helper to display an explicit error when user JH such as: "notebook_name is required when using JupyterHub"

jlewi:
@wbuchwalter Any update on this?

wbuchwalter:
@jlewi Still not sure about this one. I would like to avoid having a ton of different builders if we can converge on one.  
@ashahab, as far as I know your use case is the only one requiring this. I would like confirmation that the append builder definitly cannot suit your needs.
If not, then would there be any drawback to publish this builder as an external package? Having this on an external repo and as an additional pip package?

ashahab:
Can we avoid image building with the append builder? Our issue is with
image building.

jlewi:
I think this is obsolete.
Lets reopen if still relevant.

PR #29, opened 2018-11-28
Integrate append builder, rewritten in python

r2d4:
This PR is split into 4 different commits to make it a bit easier to review

* First, remove the fairing Go subproject - it has been rewritten in python
* Add the google/containerregistry library, forked to work with python3. It provides some primitives to manipulate docker images
* Add in the fairing append builder
* Add an example

/assign @wbuchwalter

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/29)


jlewi:
/unassign @jlewi 
I'll let others take the lead on this one.

wbuchwalter:
This look really great and will dramatically improve usability, nice job.
I talked with @r2d4 offline, this is lgtm for me except for a couple of small details:
* The current base image in the builder is not public
* `notebook_file` argument for the builder must be specified currently, which does not fit with python files or standalone notebook where we can infer the file to execute directly.
* It would be nice to have `index.docker.io` as the default registry if none is provided. (i.e if i do `builders.AppendBuilder(repository="wbuchwalter")`) 
* The generated tarball is currently saved at `.tmp.fairing.layer.tar.gz`, can we instead save this in a temporary directory somewhere that we cleanup after the build?

Issue #28, opened 2018-11-26
Admin Config


wbuchwalter:
Admin should be able to create a `ConfigMap` that would then be pulled by fairing and act as a global configuration.
This `ConfigMap` could contain settings such as:
* Which Builders are allowed
* Which Backend is allowd (kubeflow vs k8s)
* Restriction on maximum number of parallels run 
* etc.

PR #48, opened 2018-12-27
Tfjob template config map.

ashahab:
Tfjob template provides a namespaced configmap where users can provide
default images, mounts, environments, init containers, and other pod
spec resources for the tfjob to be created from the notebook

This addresses some of the issues mentioned in #28

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/48)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 examples/*/Dockerfile
 *.bak
 .vscode
-
+.idea
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -9,4 +9,5 @@ COPY setup.py /opt/fairing/
 RUN pip install -e .
 
 COPY ./ /opt/fairing
+RUN python ./setup.py test
 RUN python ./setup.py install
--- a/fairing/backend/kubernetes/manager.py
+++ b/fairing/backend/kubernetes/manager.py
@@ -7,10 +7,14 @@ from future import standard_library
 standard_library.install_aliases()
 
 import logging
+import yaml
+import json
+from jinja2 import Template
 logger = logging.getLogger(__name__)
 from pprint import pprint
 
 from kubernetes import client, config, watch
+from kubernetes.client.rest import ApiException
 from fairing.utils import is_running_in_k8s
 
 MAX_STREAM_BYTES = 1024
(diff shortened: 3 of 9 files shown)


ashahab:
@r2d4 @wbuchwalter @jlewi Can one of you mark this ok-to-test? I fixed all the unit tests in this PR.

jlewi:
/ok-to-test
If you send me a PR adding yourself to 
https://github.com/kubeflow/internal-acls/blob/master/github-orgs/kubeflow/org.yaml
I will send you an invite to the GitHub org so you can trigger tests.

@ashahab Did you consider using an admission controller to supply default values for TFJobs?

ashahab:
@jlewi I created the PR for adding me to kubeflow: https://github.com/kubeflow/internal-acls/pull/38

I haven't thought of admission controllers for defaults. I think these defaults for us would be per-namespace(corresponding to a particular ML org/team) and often would be managed by engineers not familiar with k8s, and therefore configmap is an easy way to configure.

jlewi:
@ashahab This problem of letting administrators define suitable defaults for different types of resources (e.g. TFJob / TFServing etc...) has come up before e.g. kubeflow/kubeflow#732.

I think AdmisionControllers can provide a more flexible pattern for doing this.

As an admin you can define one or more admission controllers. These controllers provide a hook that will modify the resource before it is actually created in the APIServer. This makes it easy for admins to inject common config like environment variables, mounts, etc...

I think this can be done selectively; i.e. you can configure an admission controller to only work on certain namespaces or objects with certain labels / annotations

An advantage of an admission controller is its not tied to how a user submits the resource. So it wouldn't be fairing specific. It could apply to jobs created outside of fairing.

So could we generalize this PR into something that could be used as a [web hook admission controller](https://kubernetes.io/docs/reference/access-authn-authz/webhook/) and make it easy for admins to define common config?

ashahab:
How would we configure the admission controller? We'll need something easily confgurable like the PodPreset controller provides PodPresets

jlewi:
@ashahab I think it could be configured pretty much like what you have here; i.e. using a config map to provide the web hook the default values to set.

jlewi:
You can always find me on slack if you want to have a higher bandwidth conversation.

It would be good to get @r2d4's thoughts but I think he's out.

abhi-g:
I am closing this as its been abandoned since late last year. Please reopen if needed.

PR #50, opened 2019-01-04
avoid launching both chief and worker pods when running a non distributed job

hamedhsn:
when triggering a non distributed job in kubeflow it will create two pods(1 worker, 1 chief). 
This will fix that.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/50)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setuptools.setup(
         'notebook==5.6.0',
         'jupyter==1.0.0',
         'numpy==1.15.0',
-        'kubernetes==6.0.0',
+        'kubernetes==8.0.1',
         'future==0.17.1',
         'six==1.11.0',
         'httplib2==0.12.0',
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -29,7 +29,7 @@
         'notebook==5.6.0',
         'jupyter==1.0.0',
         'numpy==1.15.0',
-        'kubernetes==6.0.0',
+        'kubernetes==8.0.1',
```
hamedhsn: The client added support for Heptio authentication for EKS(AWS)

r2d4:
@hamedhsn you'll need to sign the CLA

hamedhsn:
@jlewi I checked both emails and username all matched. not sure where the problem is tho.

hamedhsn:
@jlewi looks good now 👍

PR #52, opened 2019-01-10
Adding more optional parameters

hamedhsn:
Provide a capability to pass a few options when running job
- job_name
- image_pull_policy
- restart_policy

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/52)

### DIFF
--- a/fairing/builders/docker_builder.py
+++ b/fairing/builders/docker_builder.py
@@ -25,24 +25,29 @@ DEFAULT_IMAGE_NAME = 'fairing-job'
 class DockerBuilder(BuilderInterface):
     """A builder using the local Docker client"""
 
-    def __init__(self, 
-                repository, 
-                image_name=DEFAULT_IMAGE_NAME, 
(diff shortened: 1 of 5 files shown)


wbuchwalter:
Do we really want this to be controlled at the user level?
Fairing shouldn't expose all possible parameters that are available on the pod/job spec.
Could this be just solved by sane default in the builder?
If there is really a need for customization, could this be solved with the admin config (https://github.com/kubeflow/fairing/issues/28) instead?

hamedhsn:
hmm.. I see.
configmap make sense for really admin stuff like image_pull_policy or restart policy, however, users might want to pick the name for their jobs. This will make it easier once multiple users submit different jobs. Based on my observation, in most of the case names are chosen meaningful to the experiments.
Any further operations make it easier once it has a custom name.

hamedhsn:
closing this as it is old and was before major refactoring.

PR #57, opened 2019-01-11
logs: decode utf output

r2d4:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/57)


fairing/backend/kubernetes/manager.py
```diff
@@ -114,6 +114,6 @@ def log(self, name, namespace, extra_label_selector=None):
         if tail:
             try:
                 for chunk in tail.stream(MAX_STREAM_BYTES):                    
-                    print(chunk.rstrip())
+                    print(chunk.rstrip().decode('utf-8'))
```
jlewi: Should we be using logging module and not print?

Issue #47, opened 2018-12-17
Fire off K8s job to train python models


jlewi:
fairing should support firing off K8s jobs to train python models (e.g. xgboost/scikits) learn models.

Need to figure out
  1. What the syntax is
  1. How to move data back and forth

ashahab:
Users are bringing data to notebook. Can the same work for job?

wbuchwalter:
> If the notebook can pull in the data, why won't the job be able to do it also?

If the logic to access the data is in your code, then yes it will work out of the box.

However I suspect many users will want to access their data not through an API but instead rely on some pre-configured data store such as en EBS/AzureDisk or some already created PV. And we would need to find a way to allow users to specify these kind of sources without introducing too much complexity.

ashahab:
I see. It's about allowing users input information that's not notebook
code, but related.

jlewi:
@ashahab I think you are correct in that we want to use fairing to fire off scikit jobs as K8s jobs so they can run asynchronously on other nodes.

I think what @wbuchwalter is saying that we don't want to use fairing to move data around. To move data around we would use a shared filesystem that can be accessed by both the notebook and the K8s job. For this there are two possibilities

   1. An external shared filesystem that is directly accessed (e.g. object store like S3/HDFS/GCS)
   1. A PVC that can be mounted on the notebook pods and K8s Job

In either case its just about configuring the pods (notebook and K8s job) so that code can access the data.

To @wbuchwalter's point I think this where our discussion about using Admission Controllers could be useful. An administrator could configure one or more admission controllers so that appropriate volumes are mounted on pods to make data accessible. Or in the case of GCS/S3/HDFS the admission controller could ensure appropriate credentials are provided to each pod.

zhenghuiwang:
The xgboost example for housing works right now after several issues being fixed. IIRC, Matt has been working on making it run in notebook @r2d4

jlewi:
Will be fixed by #64

r2d4:
Fixed by #64 

/close

Issue #42, opened 2018-12-17
Train TF locally within notebook


jlewi:
Related to #39 don't use decorators

We'd like the user to be able to train the model locally within the notebook. Right now this is possible but it requires setting appropriate function decorators.

We'd like this to be more natural e.g.

```
model = MyModel()
model.train()
```

r2d4:
Fixed by #64 
/close

PR #64, opened 2019-01-29
Refactor

r2d4:
This is a pretty large refactor so let me describe some of what is going on:

* Stages were made a bit more explicit - they are preprocess, build, and deploy.
* Decorators are gone and replaced with a config builder approach
* The fairing runtime dependency is removed, so any base image can be used

### preprocess 
takes a set of inputs and translate them to a context for the builder.

**base_preprocessor** - adds the fairing runtime files and all of the files in the directory. Defaults to a python entrypoint and command.

**full_notebook_preprocessor** - schedules an entire notebook for execution

**converted_notebook_preprocessor** - converts a notebook to a python file prepares a context for that

### build
build is relatively unchanged, except many of the duplicate code has been centralized in the base_builder class

* Knative dependency is removed - we now invoke the in-cluster builder directly instead of going through a knative template and CRD. In the future I'd like to support multiple context sources like skaffold - but for now it only supports the upload-to-gcs flow.

### deploy
deploy also is relatively unchanged, and there is probably more code de-duping to do

## Config
The config is moved away from decorators, and uses a string to class map to avoid imports directly in the user's file. There have been some minor renamings in the cross over from decorator to config.

## No runtime dependency on fairing

This is accomplished by adding a runtime_config shim file that passes through to the model's train function. We automatically add this to the context, and any base image can be used.

TODO
- [ ] Update documentation
- [ ] Add some tests

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/64)


r2d4:
Fixes #47 
Fixes #42

jlewi:
/lgtm
/hold

@r2d4 None of my changes are blocking. I think the main comment I have is to sync up with @karthikv2k and make sure he's on board with the syntax and other user facing design choices. You can do that either before or after this PR is merge.d

karthikv2k:
We are meeting today to talk about syntax and other design choices and get an alignment.

PR #67, opened 2019-02-14
Add push switch to builders

r2d4:
Add an option to skip pushing the docker image for testing or local clusters like minikube or docker for desktop.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/67)

### DIFF
--- a/fairing/builders/cluster/gcs_context.py
+++ b/fairing/builders/cluster/gcs_context.py
@@ -54,7 +54,12 @@ class GCSContextSource(ContextSourceInterface):
             self.created_secret.metadata.namespace,
             client.V1DeleteOptions())
 
-    def generate_pod_spec(self, image_name):
+    def generate_pod_spec(self, image_name, push):
+        args = ["--dockerfile=Dockerfile",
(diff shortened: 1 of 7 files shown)


fairing/builders/cluster/gcs_context.py
```diff
@@ -54,7 +54,12 @@ def cleanup(self):
             self.created_secret.metadata.namespace,
             client.V1DeleteOptions())
 
-    def generate_pod_spec(self, image_name):
+    def generate_pod_spec(self, image_name, push):
+        args = ["--dockerfile=Dockerfile",
+                          "--destination=" + image_name,
+                          "--context=" + self.uploaded_context_url]
+        if not push:
+            args.append("--no-push")
```
karthikv2k: For no push case for cluster builder, where does the built image live? I am not able understand a valid use case for this.

r2d4: Testing/parity with other builders. It’s mainly functional for the docker daemon alone right now, but the append builder and cluster should optionally perform a docker load to have the same effect

PR #68, opened 2019-02-14
Add a basic function preprocessor

r2d4:
Needs #67 first

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/68)

### DIFF
--- a/examples/simple/main.py
+++ b/examples/simple/main.py
@@ -11,5 +11,5 @@ def train():
     print('Hostname: ', sess.run(hostname).decode('utf-8'))
 
 if __name__ == '__main__':
-    train = fairing.config.fn(train)
-    train()
+    remote_train = fairing.config.fn(train)
+    remote_train()
--- a/examples/simple/main.py
+++ b/examples/simple/main.py
@@ -2,11 +2,14 @@ import os
 import fairing
 import tensorflow as tf
 
-class SimpleModel():
-    def train(self):
-        hostname = tf.constant(os.environ['HOSTNAME'])
-        sess = tf.Session()
-        print('Hostname: ', sess.run(hostname).decode('utf-8'))
+fairing.config.set_builder('docker', push=False)
+fairing.config.set_deployer('job')
 
-fairing.config.set_model(SimpleModel())
-fairing.config.run()
+def train():
+    hostname = tf.constant(os.environ['HOSTNAME'])
+    sess = tf.Session()
+    print('Hostname: ', sess.run(hostname).decode('utf-8'))
+
+if __name__ == '__main__':
+    train = fairing.config.fn(train)
+    train()
--- a/fairing/config.py
+++ b/fairing/config.py
@@ -1,6 +1,7 @@
 from fairing.preprocessors.base import BasePreProcessor
 from fairing.preprocessors.converted_notebook import ConvertNotebookPreprocessor
 from fairing.preprocessors.full_notebook import FullNotebookPreProcessor
+from fairing.preprocessors.function import FunctionPreProcessor
 
 from fairing.builders.append.append import AppendBuilder
 from fairing.builders.docker.docker import DockerBuilder
@@ -24,6 +25,7 @@ preprocessor_map = {
     'python': BasePreProcessor,
     'notebook': ConvertNotebookPreprocessor,
     'full_notebook': FullNotebookPreProcessor,
+    'function': FunctionPreProcessor,
 }
 
 builder_map = {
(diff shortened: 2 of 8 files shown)


karthikv2k:
Can you please add a test to cover function preprocessor?

r2d4:
There isn't any testing infra yet so let's tackle that in a separate PR

fairing/config.py
```diff
@@ -40,68 +42,55 @@
...
-        self.get_deployer().deploy(pod_spec)
-        self.reset()
-
-    def reset(self):
-        self._builder = None
-        self._deployer = None
-        self._model = None
-        self._preprocessor = None
-
+        preprocessor = self.get_preprocessor()
```
karthikv2k: Non-blocker: we should think about moving the run method outside of config and keep the config clean of any actions.

fairing/config.py
```diff
@@ -40,68 +42,55 @@
...
-        if self._deployer is None:
-            self.set_deployer()
-        return self._deployer
-        
-    def set_model(self, model):
-        self._model = model
-    
-    def get_model(self):
-        return self._model
+        fn = deployer_map.get(self._deployer_name)
```
karthikv2k: Non-blocker: we should add an option for users to give the deployer/preprocessor/builder class directly that allows these to be in out-tree.

fairing/deployers/job/job.py
```diff
@@ -22,7 +22,7 @@ class Job(DeployerInterface):
             will generate multiple jobs.
     """
 
-    def __init__(self, namespace=None, runs=1, output=None, labels={'fairing-deployer': 'job'}):
+    def __init__(self, namespace=None, runs=1, output=None, cleanup=True, labels={'fairing-deployer': 'job'}):
```
karthikv2k: doc strings?

fairing/preprocessors/function.py
```diff
@@ -0,0 +1,43 @@
...
+                 executable=None,
+                 path_prefix=constants.DEFAULT_DEST_PREFIX,
+                 output_map={}):
+
+        super().__init__(
+            output_map=output_map,
+            path_prefix=path_prefix)
+
+
+        fairing_dir = os.path.dirname(fairing.__file__)
```
karthikv2k: Non-blocker: are these steps necessary because the docker's python runtime doesn't have fairing installed? if so, my comment is that it is better to match exactly between the user's local python env and  the remote docker python env. We can discuss this in next issue/PR.

r2d4: Yes it’s to match the target runtime to fairing shim

fairing/config.py
r2d4: Agreed.

fairing/config.py
r2d4: Yep. Run/config should be separated. They are together right now because the runtime shim will need both implementations so it’s easier to ship it as one.

examples/simple/main.py
```diff
@@ -2,11 +2,14 @@
...
-fairing.config.set_model(SimpleModel())
-fairing.config.run()
\ No newline at end of file
+def train():
+    hostname = tf.constant(os.environ['HOSTNAME'])
+    sess = tf.Session()
+    print('Hostname: ', sess.run(hostname).decode('utf-8'))
+
+if __name__ == '__main__':
+    train = fairing.config.fn(train)
```
karthikv2k: giving the new function the same name ("train") is confusing. Probably something like remote_train = fairing.config.fn(train); remote_train() is less confusing.

fairing/preprocessors/function.py
```diff
@@ -0,0 +1,43 @@
...
+import glob
+import os
+from pathlib import Path
+
+from fairing.constants import constants
+from .base import BasePreProcessor
+
+FUNCTION_SHIM = 'function_shim.py'
+
+
```
karthikv2k: doc strings?

examples/simple/main.py
```diff
@@ -2,11 +2,14 @@
...
 
-fairing.config.set_model(SimpleModel())
-fairing.config.run()
\ No newline at end of file
+def train():
+    hostname = tf.constant(os.environ['HOSTNAME'])
+    sess = tf.Session()
+    print('Hostname: ', sess.run(hostname).decode('utf-8'))
+
+if __name__ == '__main__':
```
karthikv2k: can you also add an example in a notebook?

fairing/functions/function_shim.py
```diff
@@ -0,0 +1,17 @@
+import argparse
+from importlib import import_module
+
+
+def call(module_name, class_name, function_name):
+    mod = import_module(module_name, function_name)
```
karthikv2k: In the example, train function is defined in the __main__ module. How does this work if function_shim is the __main__ module when it is executed remotely.

examples/simple/main.py
r2d4: Sure. What I imagined was that the user could add/remove this line and keep the rest of their code the same to do local/remote training.

fairing/functions/function_shim.py
r2d4: function_shim calls the function directly, if that makes sense

examples/simple/main.py
r2d4: Will do

fairing/functions/function_shim.py
karthikv2k: what will be the values of module_name and function_name in the example given here?

examples/simple/main.py
r2d4: This won't work in the notebook yet - we'll need #69

fairing/preprocessors/function.py
r2d4: done

fairing/preprocessors/function.py
```diff
@@ -0,0 +1,49 @@
...
+class FunctionPreProcessor(BasePreProcessor):
+    """
+    FunctionPreProcessor preprocesses a single function.
+    It sets as the command a function_shim that calls the function directly.
+
+    args: function_name - the name of the function to be called
+    """
+    def __init__(self,
+                 function_name,
+                 executable=None,
```
karthikv2k: there is no doc strings in BasePreProcessor so I am not able to get clear idea on what executable variable is and why it is not passed to super().\_\_init\_\_ fn.

examples/simple/main.py
karthikv2k: ok. #74 can handle this use case too.

fairing/functions/function_shim.py
r2d4:
```
    Command:
      python
      /app/function_shim.py
      --module_name
      main
      --function_name
      train
```

fairing/preprocessors/function.py
r2d4: good point - removed it and added a docstring in the base preprocessor

fairing/functions/function_shim.py
karthikv2k: looks good.

PR #77, opened 2019-02-21
Add Travis setup and Fairing config tests

### LINKED ISSUE #71: Setup a basic CI/CD system
Let's use the basic version of travis ci to get unit tests and simple integration test using local docker builder.

rpasricha:
Tests are passing on my personal repo: https://travis-ci.com/rpasricha/fairing/

Known issue:
Many classes (e.g. deployers) assume a Kubeconfig is present, which cause Travis tests to fail. I updated the cluster builder but other classes will need to be updated if they should be accessible without an existing Kubeconfig set up.

Fixes #71

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/77)


.travis.yml
```diff
@@ -0,0 +1,11 @@
+language: python
+# Python versions to test
+python:
+    - "3.5"
+# Command to install dependencies
+install:
+    - pip install -r requirements.txt
```
karthikv2k: do we need this step as the requirements are installed as part of setup.py?

rpasricha: True, I'll go ahead and remove this line

Issue #73, opened 2019-02-15
Add support for managed backends


rpasricha:
In addition to supporting Kubeflow and native Kubernetes job submissions, we should provide the option for users to swap out their deployment for a managed service as well.

Issue #65, opened 2019-02-05
Builder should return just the image


r2d4:
Instead of a podspec. The container entrypoint and environment variables should be baked into the container manifest.

/area fairing
/priority p1

PR #86, opened 2019-02-26
Add GCP deployer

rpasricha:
#73 

- Update Append builder to add command, user, and environment variables to the image manifest directly.
- Add GCPJob deployment option.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/86)


fairing/deployers/gcp/gcp.py
```diff
@@ -0,0 +1,56 @@
...
+
+        auth = subprocess.run(['gcloud', 'auth', 'print-access-token'],
+                              stdout=subprocess.PIPE).stdout[:-1].decode('utf-8')
+        headers = {
+            'Content-Type': 'application/json',
+            'Authorization': 'Bearer {}'.format(auth)
+        }
+
+        try:
+            r = requests.post('https://ml.googleapis.com/v1/projects/{}/jobs'.format(self._project_id),
```
r2d4: Why aren't you using the GCP python library?

fairing/builders/append/append.py
```diff
@@ -73,7 +74,10 @@ def _build(self, transport, src):
         creds = docker_creds.DefaultKeychain.Resolve(src)
         with v2_2_image.FromRegistry(src, creds, transport) as src_image:
             with open(self.context_file, 'rb') as f:
-                new_img = append.Layer(src_image, f.read())
+                new_img = append.Layer(src_image, f.read(),
```
r2d4: Can you add this to the dockerfile generator as well? Or open up an issue to do it?

fairing/config.py
```diff
@@ -36,7 +37,8 @@
 
 deployer_map = {
     'job': Job,
-    'tfjob': TfJob
+    'tfjob': TfJob,
+    'gcp': GCPJob
```
r2d4: GCP the right name? Or CMLE?

fairing/deployers/gcp/gcp.py
rpasricha: Good catch, updated to use the Python client library

fairing/config.py
rpasricha: I used GCP to keep it general as possible as a managed deployment option. What are your thoughts?

fairing/builders/append/append.py
rpasricha: Sure, I'll create an issue for this

rpasricha: Actually, looks like one already exists -- #65

fairing/config.py
karthikv2k: LGTM

Issue #46, opened 2018-12-17
support local prediction with python models (scikits/xgboost)


jlewi:
We need to support running prediction inside the notebook for python models (e.g. scikits/xgboost).

What should the input output format be? 

Should we use numpy arrays?
Should we follow Seldon's API?

/cc @cliveseldon

r2d4:
/assign @r2d4
for documenting/exploring xgboost example.

jlewi:
@r2d4 @karthikv2k How's this coming? What work is remaining after #91?

karthikv2k:
Needs #95 [WIP] to support deploying models from Notebook.

PR #91, opened 2019-02-28
Add Serving Deployer

r2d4:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/91)


r2d4:
Fixes https://github.com/kubeflow/fairing/issues/46

r2d4:
Example is taken from https://github.com/kubeflow/examples/tree/master/xgboost_ames_housing

examples/prediction/HousingServe.py
```diff
@@ -0,0 +1,42 @@
...
+import numpy as np
+
+MODEL_FILE = 'trained_ames_model.dat'
+
+fairing.config.set_builder('docker',
+registry='gcr.io/mrick-gcp',
+base_image="seldonio/seldon-core-s2i-python3:0.4")
+
+fairing.config.set_deployer('serving', serving_class="HousingServe")
+fairing.config.run()
```
rpasricha: (Not blocking) Would the location of `fairing.config.run()` change depending on the preprocessor that's used? It's a bit odd to see it before the class definition.

fairing/deployers/serving/serving.py
```diff
@@ -0,0 +1,82 @@
...
+            logger.warn(json.dumps(job_output))
+            service_output = api.sanitize_for_serialization(self.service_spec)
+            logger.warn(json.dumps(service_output))
+
+        v1_api = k8s_client.CoreV1Api()
+        apps_v1 = k8s_client.AppsV1Api()
+        self.deployment = apps_v1.create_namespaced_deployment(self.namespace, self.deployment_spec)
+        self.service = v1_api.create_namespaced_service(self.namespace, self.service_spec)
+
+        logger.warn("Training job {} launched.".format(self.deployment.metadata.name))
```
rpasricha: Replace with "endpoint launched"

fairing/deployers/serving/serving.py
```diff
@@ -0,0 +1,82 @@
...
+    Serves a prediction endpoint using Kubernetes deployments and services 
+    
+    serving_class: the name of the class that holds the predict function.
+    
+    """
+    def __init__(self, serving_class, namespace=None, runs=1, ):
+        super(Serving, self).__init__(namespace, runs)
+        self.serving_class = serving_class
+
+    def create_resource(self):
```
rpasricha: Do we need this in the Serving deployer? I don't see where it's being used.

examples/prediction/HousingServe.py
r2d4: Nope. I put it here because since it running everything remotely, its faster to put it closer to the top (so heavy libraries arent loaded before it gets ran). But in reality it can be any where

fairing/deployers/serving/serving.py
r2d4: good catch - not used. deleted

fairing/deployers/serving/serving.py
r2d4: done

examples/prediction/HousingServe.py
karthikv2k: It was confusing for me too. I think just for clarity we can place it at the end of the file.

Issue #83, opened 2019-02-26
Evaluate Cloud Build as a CI/CD system


rpasricha:
To make it easier to access various services and run end-to-end and integration tests.

PR #92, opened 2019-03-01
Add cloud build Dockerfile and build steps

rpasricha:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/92)

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,25 @@
+FROM python:3.5
+
+WORKDIR /opt/fairing
+
+# Install dependencies
+COPY setup.py requirements.txt /opt/fairing/
+RUN pip install -e .
+
+# Install gcloud
+RUN mkdir -p /builder && \
+	wget -qO- https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz | tar zxv -C /builder && \
+	CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
+		--bash-completion=false \
+		--disable-installation-options && \
+	rm -rf ~/.config/gcloud
+ENV PATH=/builder/google-cloud-sdk/bin/:$PATH
+
+# Install test dependencies
+RUN pip install pytest
+
+# Install Fairing
+COPY . /opt/fairing
+RUN python setup.py install
+
+CMD pytest
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,22 +1,24 @@
 FROM python:3.5
 
 WORKDIR /opt/fairing
+ARG CLOUD_SDK_VERSION="236.0.0"
 
-# Install dependencies
-COPY setup.py requirements.txt /opt/fairing/
-RUN pip install -e .
-
-# Install gcloud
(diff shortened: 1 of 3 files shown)


rpasricha:
#83

Dockerfile.test
```diff
@@ -0,0 +1,25 @@
+FROM python:3.5
+
+WORKDIR /opt/fairing
+
+# Install dependencies
+COPY setup.py requirements.txt /opt/fairing/
```
karthikv2k: is this step redundant? line 23 would install deps too.

Dockerfile.test
```diff
@@ -0,0 +1,25 @@
...
+ENV PATH=/builder/google-cloud-sdk/bin/:$PATH
+
+# Install test dependencies
+RUN pip install pytest
+
+# Install Fairing
+COPY . /opt/fairing
+RUN python setup.py install
+
+CMD pytest
```
karthikv2k: doesn't tests typically run with tests as working dir?

Dockerfile.test
r2d4: No, this copies over the requirement files first to download the dependencies for more optimal caching. 

If done together with line 23, then every time the source changes, we'd redownload all the dependencies

Dockerfile.test
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /opt/fairing
+
+# Install dependencies
+COPY setup.py requirements.txt /opt/fairing/
+RUN pip install -e .
+
+# Install gcloud
+RUN mkdir -p /builder && \
+	wget -qO- https://dl.google.com/dl/cloudsdk/release/google-cloud-sdk.tar.gz | tar zxv -C /builder && \
+	CLOUDSDK_PYTHON="python2.7" /builder/google-cloud-sdk/install.sh --usage-reporting=false \
```
r2d4: - python 3.5?
- pin to a specific gcloud version? 
- move this to the very top of the dockerfile, after WORKDIR

Dockerfile outline:
1. install gcloud
2. install pytest
3. copy setup.py and requirements, run pip install
4. copy over source
5. install fairing

Dockerfile.test
rpasricha: Yes, this helps us cache the dependencies in a different layer. The base image is not stored by default though, we'd need to add a separate step to pull the last image and use that as a cache - I'll go ahead and add that.

https://cloud.google.com/cloud-build/docs/speeding-up-builds

Dockerfile.test
rpasricha: pytest will extract tests from subdirectories, so runs successfully from the root project dir as well.

Dockerfile.test
rpasricha: Updated

Dockerfile.test
rpasricha: Updated to cache test image across builds

rpasricha:
> For optimal caching:
> 
> Dockerfile outline:
> 
> 1. install gcloud
> 2. install pytest
> 3. copy setup.py and requirements, run pip install
> 4. copy over source
> 5. install fairing

Updated Dockerfile to have this order, thanks!

PR #96, opened 2019-03-04
explicitly set restart_policy to never for job

r2d4:
Job enforces that the pod template spec not have restart_policy set to 'Always' which is the default value. Deployment, however, enforces that restart_policy is not 'Never', so it must be set in the deployer. 

Temporary workaround until we create the entire spec in the deployer.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/96)


r2d4:
ref #65

PR #103, opened 2019-03-06
Check additional locations when retrieving project_id and credentials

rpasricha:
Currently we are just using a user-provided credentials file or checking the `GOOGLE_APPLICATION_CREDENTIALS` environment variable.

This adds additional checks, e.g. gcloud, allowing authentication with end-user credentials as well.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/103)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 # Command to install dependencies
 install:
     - python setup.py install
+    - pip install --upgrade pytest
 # Command to run tests
 script:
     - pytest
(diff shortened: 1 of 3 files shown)


karthikv2k:
Can you add unit tests?
/hold

rpasricha:
Added unit tests, and upgraded pytest in the travis config file as the tests used a newer feature of pytest (the tmp_path argument) that was not included in the default version.

PR #112, opened 2019-03-11
Kubeflow credentials, tarfile fixes, unit tests

### LINKED ISSUE #101: Docker pull fails for appendbuilder with virtualenv
If there is a virtualenv in the same directory as my code, it gets packaged and added to the image when using the append builder. However, this image is broken and a `docker pull` will fail with the following error:

```
docker: failed to register layer: Error processing tar file(duplicates of file paths not supported): .
See 'docker run --help'.
```

### LINKED ISSUE #106: Submitted kubeflow jobs don't have access to other services
E.g. GCS -- we need to make sure that jobs that are submitted have the appropriate kubeflow-user service account credentials set.

/kind bug

rpasricha:
A few combined changes in this PR
* Separate tests into unit and integration (only run Travis for unit tests)
* Authenticate with Kubeflow cluster in Cloud Build to run integration tests
* Add integration test with Kubernetes job deployment
* Mount Kubeflow service account credentials in TFJob (Fixes #106)
* Skip duplicate file additions when generating context tarfile (Fixes #101)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/112)


rpasricha:
I'm getting the following test failure for the integration test:

```
  File "/app/function_shim.py", line 15, in <module>
    call(args.serialized_fn...ion_shim.py", line 6, in call
    fn = cloudpickle.load(f)
ImportError: No module named \'test_kubeflow_training\'
```

@r2d4 @karthikv2k Any ideas?

karthikv2k:
this is due to the function not in __main__ module. We should set the fn.__module__ = "__main__" in the function preprocessor before serialization. This is not an issue when users are working in the notebooks as all the things defined in a notebook is in __main__ module. This issue mainly occurs while testing.

rpasricha:
Sounds good, I'll update the module of the object in the integration test.

rpasricha:
Added main module, and fixed another test failure.

The default preprocessor `output_map` argument was an empty dictionary, which is mutable, so file paths from one preprocess call were getting persisted to the next. Updated to use `None` as the default instead of an empty dict.

fairing/deployers/tfjob/tfjob.py
```diff
@@ -23,6 +23,31 @@ def generate_deployment_spec(self, pod_template_spec):
         """Returns a TFJob template"""
         self.set_container_name(pod_template_spec)
 
+        # Set appropriate secrets and volumes to enable kubeflow-user service
+        # account.
+        env_var = k8s_client.V1EnvVar(
```
hamedhsn: This will break in other platforms than gcp? cant this be optional?
same for volume_mount

r2d4: I agree, we'll need to add this as something configurable in the deployer

karthikv2k: we should probably have a global config that sets the deployment is happening in GCP and necessary things like gcloud for the builder and auth setup for the deployers are automatically configured.

fairing/preprocessors/base.py
```diff
@@ -61,9 +61,13 @@ def context_map(self):
 
     def context_tar_gz(self, output_file=constants.DEFAULT_CONTEXT_FILENAME):
         self.input_files = self.preprocess()
-        with tarfile.open(output_file, "w:gz") as tar:
+        added_files = set()
```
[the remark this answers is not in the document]
karthikv2k: I see it is used in line 68.

tests/integration/test_kubeflow_training.py
```diff
@@ -0,0 +1,32 @@
...
+
+    remote_train = fairing.config.fn(train_fn)
+
+    # Capture stdout
+    sys.stdout = io.StringIO()
+    remote_train()
+    output = sys.stdout.getvalue()
+    sys.stdout = sys.__stdout__
+
+    assert 'train_fn' in output
```
karthikv2k: it is better to have printed value other than the function name. I can think of instances where the error message/stack trace have the function name that will give a false test result.

fairing/preprocessors/base.py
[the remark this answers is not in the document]
r2d4: ack

fairing/deployers/tfjob/tfjob.py
hamedhsn: global config sets for deployer make sense. then we could also add other configurations like different custom labels for different jobs, volume, gpu setting etc.

r2d4: Right now tfjob superclasses job so maybe we just have an actual base deployer like we do for builders

rpasricha: Right, thanks for the catch! In order to keep this PR manageable, I'll plan to add a flag to the TFJob deployer, and file a FR for the global config sets.

With the config sets it'll likely be useful to have a base deployer as well, in order for changes like these to be shared accordingly.

fairing/preprocessors/base.py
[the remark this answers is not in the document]
rpasricha: Yes, I found a quite a few places where duplicate files can be added to the tarfile, so this is a check to make sure that duplicates are skipped when we are creating the context package itself.

The context map also means that it is possible for multiple input files to map to a single file in the archive, so we store the set of filenames that have actually been added.

tests/integration/test_kubeflow_training.py
rpasricha: Sure, I'll update the output to use a different string

fairing/preprocessors/base.py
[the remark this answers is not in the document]
r2d4: Can we move this to be a set in the init function? It's a bit weird to have it be a map, but enforce set qualities later on.

fairing/deployers/tfjob/tfjob.py
rpasricha: Created #118

fairing/preprocessors/base.py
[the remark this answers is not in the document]
rpasricha: That sounds reasonable, so we can essentially merge `context_map` and `added_files` into a single set. This will also help to keep the source / destination filenames consistent.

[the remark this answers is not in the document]
rpasricha: Updated to reverse the context map direction to avoid duplicates (destination -> source file), and also add warning messages when duplicate files are skipped.

PR #113, opened 2019-03-14
add custom labels to deployer

hamedhsn:
+ options to add custom labels for job
+ fixing the fairing-deployer label. right now it always set it to job
+ missing functions in runtime_config.py(crashing the job right now)

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/113)

### DIFF
--- a/fairing/deployers/job/job.py
+++ b/fairing/deployers/job/job.py
@@ -11,7 +11,7 @@ from fairing.deployers.deployer import DeployerInterface
 
 logger = logging.getLogger(__name__)
 DEFAULT_JOB_NAME = 'fairing-job-'
-DEFAULT_LABELS = {'fairing-deployer': 'job'}
+DEPLOPYER_TYPE = 'job'
 
 
 class Job(DeployerInterface):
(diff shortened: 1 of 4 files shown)


fairing/runtime_config.py
```diff
@@ -23,6 +23,12 @@ def set_deployer(self, name, **kwargs):
     def get_deployer(self, **kwargs):
         return self._deployer
 
+    def get_model(self):
```
r2d4: These aren't necessary anymore. What crash were you seeing?

hamedhsn: hmm I think it confuses the name. because we have config module then inside the config you instantiate the Config() class and RuntimeConfig to `config`. and when call fairing.config I think the second one is called.
I am running a tfjob from example main folder:
```➜  ks_app git:(master) ✗ kl mljob-kubeflow-t4r74-worker-0
  File "/app/main.py", line 90, in <module>
    fairing.config.set_model(TensorflowModel())
AttributeError: 'RuntimeConfig' object has no attribute 'set_model'```

r2d4: Ah set_model should be removed from runtime config as well. Sent a PR. You only need to call fairing.run, not set_model

https://github.com/kubeflow/fairing/pull/114

hamedhsn: k. I removed them as well.

PR #117, opened 2019-03-14
Update examples

rpasricha:
Updated examples to use the current API / function preprocessor. Updated in the following directories:

* `kubeflow-gke/`
* `kubernetes/`
* `notebook/`
* `simple/`

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/117)


examples/kubeflow-gke/main.py
```diff
@@ -43,43 +43,42 @@
...
+        if step % 100 == 0:
+            print("At step {}, loss = {}".format(step, loss_value))
+            summary_str = sess.run(summary, feed_dict=feed_dict)
+            summary_writer.add_summary(summary_str, step)
+            summary_writer.flush()
 
 if __name__ == '__main__':
-    fairing.config.set_model(TensorflowModel())
-    fairing.config.run()
+    train = fairing.config.fn(train)
```
r2d4: I'd prefer to not have most of the examples depend on the function preprocessor and just work normally. Most times it will probably cause more issues than not

karthikv2k: I'd recommend function preprocessor for notebooks only. Only in notebooks there is a need to capture the current state of the objects so makes sense to use fn preprocessor's pickling feature. For python files, just simply executing the file remotely would be better without much restriction.

karthikv2k: I see two main way of people providing execution unit:
1. Users working in notebooks or ipython shells, create some code interactively and wants to execute that in the remote system. For this fn processor with cloudpickle is the right solution.
2. Users working in python files or notebooks, write code and except linear execution of it in the remote system. For this, base preprocessor or notebook preprocessor is the right solution.

thoughts?

rpasricha: Sure, that's a reasonable separation of use cases. I'll work on updating the Python file examples to use alternative preprocessors, and keep the function preprocessor used in the notebook cases.

rpasricha:
Updated kubeflow-gke and kubernetes examples to use fairing.config.run instead of function preprocessor.

PR #135, opened 2019-03-20
High level APIs for train and deploy in notebooks

karthikv2k:
This is a first attempt at creating high level APIs that hides as much of underlying details as possible. This is currently focused on notebooks with function preprocessor and we will be expanding to other preprocessor and work with python files.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/135)


karthikv2k:
Some comments from prev PR (now closed) https://github.com/kubeflow/fairing/pull/123#issuecomment-474096863

fairing/notebook/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, obj, base_docker_image, docker_registry=None, input_files=[]):
+        if not docker_registry:
+            guessed_docker_registry = guess_docker_registry()
+            if not guessed_docker_registry:
+                raise RuntimeError("Not able to guess docker registry, please specify docker_registry explicitly.")
+            self.docker_registry = guessed_docker_registry
+        else:
+            self.docker_registry = docker_registry
+        preprocessor = guess_preprocessor(obj, input_files=input_files)
+        self.builder = DockerBuilder(preprocessor=preprocessor,
```
r2d4: Suggest using the Cluster builder instead, which will work on Kubeflow notebooks

fairing/notebook/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+
+def guess_docker_registry():
+    try:
+        gcp_project = fairing.cloud.gcp.guess_project_name()
+        registry = 'gcr.io/{}'.format(gcp_project)
+        return registry
+    except Exception as e:
+        return None
+
+class BaseJob:
```
r2d4: these classes should live somewhere outside the notebook package, possibly their own

fairing/notebook/jobs.py
```diff
@@ -0,0 +1,61 @@
...
+from fairing.deployers.job.job import Job
+from fairing.deployers.serving.serving import Serving
+
+def guess_preprocessor(obj, *args, **kwargs):
+    if get_execution_obj_type(obj) != ObjectType.NOT_SUPPORTED:
+        return FunctionPreProcessor(obj, *args, **kwargs) 
+    else:
+        raise NotImplementedError("obj param should be a function or a class, got {}".format(type(obj)))
+
+def guess_docker_registry():
```
r2d4: these should be handled lower down the stack, possibly in the base builder

fairing/utils.py
```diff
@@ -33,3 +35,51 @@ def crc(file_name):
...
+    from fairing.preprocessors.function import FunctionPreProcessor
+    
+    if get_execution_obj_type(entry_point) != ObjectType.NOT_SUPPORTED:
+        return FunctionPreProcessor(entry_point, *args, **kwargs)
+    else:
+        raise NotImplementedError(
+            "obj param should be a function or a class, got {}".format(type(entry_point)))
+
+
+def guess_docker_registry():
```
r2d4: these checks belong in the actual functions that need them if that makes sense

fairing/utils.py
```diff
@@ -33,3 +35,51 @@ def crc(file_name):
...
+    try:
+        gcp_project = fairing.cloud.gcp.guess_project_name()
+        registry = 'gcr.io/{}'.format(gcp_project)
+        return registry
+    except Exception as e:
+        print(e)
+        return None
+
+
+def is_docker_daemon_exists():
```
r2d4: the utils package shouldn't have any dependencies on packages outside the stdlib

r2d4: otherwise you end up with a dependency on heavy clients like docker when you just need to generate a random UUID

fairing/utils.py
```diff
@@ -33,3 +35,51 @@ def crc(file_name):
 
 def random_tag():
     return str(uuid.uuid4()).split('-')[0]
+
+
+def guess_preprocessor(entry_point, *args, **kwargs):
+    #Trying to avoid any clyclic dependencies by importing locally within the method
+    from fairing.functions.function_shim import get_execution_obj_type, ObjectType
```
r2d4: this should be an indicator that utils is not the right package for this

fairing/notebook/jobs.py
karthikv2k: Yes cluster builder makes sense and also useful to reduce network bandwidth if the user is developing out of their notebook.

fairing/notebook/jobs.py
karthikv2k: Ack.

fairing/utils.py
karthikv2k: Removed guess_docker_registry and made docker registry required param for now. Let's address how to provide good defaults based on the environment (KF on-prem, KF on GKE, managed backend, etc)

fairing/utils.py
karthikv2k: Ack. Moved util functions out of the common utils module.

fairing/utils.py
karthikv2k: Ack.

fairing/notebook/jobs.py
karthikv2k: Removed guess_docker_registry and made docker registry required param for now. Let's address how to provide good defaults based on the environment (KF on-prem, KF on GKE, managed backend, etc)

fairing/ml_tasks/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+    Base class for handling high level ML tasks.
+    args:
+        entry_point: An object or reference to the source code that has to be deployed.
+        base_docker_image: Name of the base docker image that should be used as a base image
+            when building a new docker image as part of an ML task deployment.
+        docker_registry: Docker registry to store output docker images.
+        input_files: list of files that needs to be packaged along with the entry point.
+            E.g. local python modules, trained model weigths, etc.
+    """
+    def __init__(self, entry_point, base_docker_image, docker_registry, input_files=[]):
```
rpasricha: `input_files` should be None instead of an empty list -- mutable default args will persist across method calls.

fairing/ml_tasks/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+        gussed_builder = guess_builder(needs_deps_installation=True)
+        logger.warn("Using builder: {}".format(gussed_builder.__name__))
+
+        self.builder = gussed_builder(preprocessor=preprocessor,
+                                      base_image=base_docker_image,
+                                      registry=self.docker_registry)
+
+    def _build(self):
+        self.builder.build()
+        self.pod_spec = self.builder.generate_pod_spec()
```
rpasricha: How will pod_spec_mutators be handled in the high level API?

fairing/ml_tasks/tasks.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class TrainJob(BaseTask):
+
+    def __init__(self, entry_point, base_docker_image, docker_registry, input_files=[]):
+        super().__init__(entry_point, base_docker_image, docker_registry, input_files)
+
+    def submit(self):
+        self._build()
+        deployer = Job()
```
rpasricha: Should we prefer TFJob over the standard Job deployer? Doing so will help with KF UI integration, etc.

fairing/ml_tasks/tasks.py
karthikv2k: Ack. Good catch.

fairing/ml_tasks/tasks.py
karthikv2k: I am thinking about how to enable more granular configuration like pod_spec_mutators when users are using these high level APIs. Couple of approach I am thinking: users can use fairing.config.set_<>(..) to set any granular configuration or a backend class (Kubeflow, GCP, etc) can set these good defaults or combination of both. But to keep this PR manageable, let's address that in future PRs. This topic needs more discussion.

fairing/ml_tasks/tasks.py
karthikv2k: I prefer K8's Job because it is a stable API and sufficient for single node training. Also it let's users with only K8s deployment try Fairing before jumping into KF.

rpasricha: Okay, that makes sense

rpasricha:
/lgtm

Will let @r2d4 review and approve as well.

PR #136, opened 2019-03-20
Refactor Service Account Credentials

### LINKED ISSUE #106: Submitted kubeflow jobs don't have access to other services

rpasricha:
Fixes #106 

Move credential mounting to Kubernetes manager and update both Job and TFJob deployers to mount credentials if desired.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/136)

### DIFF
--- a/fairing/deployers/job/job.py
+++ b/fairing/deployers/job/job.py
@@ -26,7 +26,7 @@ class Job(DeployerInterface):
     def __init__(self, namespace=None, runs=1, output=None,
                  cleanup=True, labels=None, job_name=DEFAULT_JOB_NAME,
                  stream_log=True, deployer_type=DEPLOPYER_TYPE,
-                 mount_credentials=False):
+                 pod_spec_mutators=None):
         if namespace is None:
             self.namespace = utils.get_default_target_namespace()
         else:
@@ -41,7 +41,7 @@ class Job(DeployerInterface):
         self.cleanup = cleanup
         self.stream_log = stream_log
         self.set_labels(labels, deployer_type)
-        self.mount_credentials = mount_credentials
+        self.pod_spec_mutators = pod_spec_mutators or []
 
     def set_labels(self, labels, deployer_type):
         self.labels = {'fairing-deployer': deployer_type}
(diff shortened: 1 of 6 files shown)


fairing/deployers/job/job.py
```diff
@@ -25,7 +25,8 @@ class Job(DeployerInterface):
 
     def __init__(self, namespace=None, runs=1, output=None,
                  cleanup=True, labels=None, job_name=DEFAULT_JOB_NAME,
-                 stream_log=True, deployer_type=DEPLOPYER_TYPE):
+                 stream_log=True, deployer_type=DEPLOPYER_TYPE,
+                 mount_credentials=False):
```
r2d4: WDYT about this being not so GCP specific?

How about a list of functions that mutates the pod spec? One such function could be `add_credentials_to_pod_spec` (rename it to `add_gcp_credentials`

So the API would look like

```
def __init__(self, ...., pod_spec_mutators=[add_gcp_credentials])
```

And then the pod spec would be ran through the mutators in order

```
for each fn in pod_mutators:
   fn(self.pod_spec)
```

The interface for the pod spec mutator would just be to accept a pod_spec and return a mutated pod spec (or return a newly modified one)

fairing/kubernetes/manager.py
```diff
@@ -56,6 +56,40 @@ def delete_deployment(self, name, namespace):
             namespace,
             client.V1DeleteOptions())
 
+    def secret_exists(self, name, namespace):
+        secrets = client.CoreV1Api().list_namespaced_secret(namespace)
+        secret_names = [secret.metadata.name for secret in secrets.items]
+        return name in secret_names
+
+    def add_credentials_to_pod_spec(self, pod_spec):
```
r2d4: This method doesn't require the kubernetes API endpoint so it doesn't have to be on the Manager class and can work by itself

fairing/deployers/job/job.py
r2d4: This will have the benefit of 

1. making it easier to plug in different clouds
2. making it easier to modify the pod spec in different ways (we'll most likely have to add more volume mounts, environment variables, and secrets in the future)

rpasricha: Thanks for the suggestion! I updated the implementation to use the `pod_spec_mutators` approach, with the deployers taking in a list of mutators accordingly. 

My only concern with this is that it's become more complex from the user-facing point of view. In particular, they need to do something like below:

```
fairing.config.set_deployer(
    'job',
    pod_spec_mutators=[fairing.cloud.gcp.add_gcp_credentials],
    namespace='kubeflow'
)
```

fairing/kubernetes/manager.py
rpasricha: Moved to `fairing/cloud/gcp.py`

Also updated the function to take as input the kubernetes manager and namespace, so we can check if the secret exists in the function itself.

rpasricha:
Also added integration tests

PR #142, opened 2019-03-26
Full notebook preprocessor updates

rpasricha:
* Normalize filepaths in base preprocessor (to avoid further duplicates)
* Add `input_files` argument to full_notebook preprocessor
* Updates to jupyter nbconvert command
    * Allow execution to continue if one cell fails with an error
    * Remove timeouts on cell execution

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/142)

### DIFF
--- a/fairing/preprocessors/full_notebook.py
+++ b/fairing/preprocessors/full_notebook.py
@@ -5,6 +5,7 @@ from fairing.notebook import notebook_util
 
 
 class FullNotebookPreProcessor(BasePreProcessor):
+    # TODO: Allow configuration of errors / timeout options
     def __init__(self,
                  notebook_file=None,
                  input_files=None,
@@ -17,7 +18,8 @@ class FullNotebookPreProcessor(BasePreProcessor):
             notebook_file = notebook_util.get_notebook_name()
 
         input_files = input_files or []
-        input_files.append(notebook_file)
+        if notebook_file not in input_files:
+            input_files.append(notebook_file)
 
         super().__init__(
             executable=notebook_file,
--- a/fairing/preprocessors/full_notebook.py
+++ b/fairing/preprocessors/full_notebook.py
@@ -7,16 +7,21 @@ from fairing.notebook import notebook_util
 class FullNotebookPreProcessor(BasePreProcessor):
     def __init__(self,
                  notebook_file=None,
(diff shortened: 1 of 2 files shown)


fairing/preprocessors/full_notebook.py
```diff
@@ -7,16 +7,21 @@
 class FullNotebookPreProcessor(BasePreProcessor):
     def __init__(self,
                  notebook_file=None,
-                 command=["jupyter", "nbconvert", "--stdout", "--to", "notebook", "--execute"],
+                 input_files=None,
+                 command=["jupyter", "nbconvert", "--stdout", "--to", "notebook", "--execute", "--allow-errors",
```
karthikv2k: I think users should be able to configure allow errors behavior. I don't see executing past an error is that common of a use case but on the other hand users may not want to run costly operations like training if any preprocessing is failed.

fairing/preprocessors/full_notebook.py
```diff
@@ -7,16 +7,21 @@
...
                  path_prefix=constants.DEFAULT_DEST_PREFIX,
                  output_map=None):
-        
+
         if notebook_file is None and notebook_util.is_in_notebook():
             notebook_file = notebook_util.get_notebook_name()
-        
+
+        input_files = input_files or []
+        input_files.append(notebook_file)
```
karthikv2k: can you check if the notebook_file is already in input_files and then add it? more general question, does input_files should be a set instead of a list? I understand that users don't use set regularly but internally we should be using set instead of list.

fairing/preprocessors/full_notebook.py
rpasricha: Added a todo for now. From testing, I've found that nbconvert does not display any output while the notebook is executing, which makes it challenging to monitor long-running jobs.

So we will want to prioritize Papermill (#33) and we can add additional configuration options there.

fairing/preprocessors/full_notebook.py
rpasricha: Added the check to make sure notebook_file is already in input_files.

The base preprocessor handles conversion of the list of input files to a set, and also normalizes relative paths to avoid duplicates. When the list of input files is passed to the BasePreProcessor constructor, the duplicates will be removed.

PR #159, opened 2019-03-29
Skip creating secret in context preparer

r2d4:
We do this in the pod_spec_mutator functions now

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/159)

### DIFF
--- a/fairing/builders/cluster/gcs_context.py
+++ b/fairing/builders/cluster/gcs_context.py
@@ -21,7 +21,6 @@ class GCSContextSource(ContextSourceInterface):
         if self.gcp_project is None:
             self.gcp_project = gcp.guess_project_name()
         self.uploaded_context_url = self.upload_context(context_filename)
-        self.created_secret = self.create_secret()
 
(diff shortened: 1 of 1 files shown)


fairing/builders/cluster/gcs_context.py
```diff
@@ -30,29 +29,9 @@ def upload_context(self, context_filename):
                     bucket_name=self.gcp_project,
                     blob_name='fairing_builds/' + context_hash,
                     file_to_upload=context_filename)
-        
-    def create_secret(self):
```
abhi-g: @r2d4 Is it the case that based on our analysis of all paths a user may take (local/KF/hosted notebooks), there won't be a need to create secret?

r2d4: The (possible) secret generation was moved into a pod_spec_mutator - the GCS context preparer shouldn't be responsible for creating this secret.

abhi-g:
you might need to self-approve

PR #171, opened 2019-04-03
Update full notebook tests and convert absolute to relative paths

rpasricha:
Fix test failures by updating tests to use Python base image, install Papermill and notebook as explicit dependencies.

Also fixed an issue where absolute notebook paths weren't getting copied to the container by converting to relative paths.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/171)

### DIFF
--- a/fairing/preprocessors/full_notebook.py
+++ b/fairing/preprocessors/full_notebook.py
@@ -23,7 +23,7 @@ class FullNotebookPreProcessor(BasePreProcessor):
 
         relative_notebook_file = notebook_file
         # Convert absolute notebook path to relative path
-        if notebook_file[0] == '/':
+        if os.path.isabs(notebook_file[0]):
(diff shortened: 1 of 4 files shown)


fairing/preprocessors/full_notebook.py
```diff
@@ -20,12 +21,17 @@ def __init__(self,
         if notebook_file is None:
             raise ValueError('A notebook_file must be provided.')
 
+        relative_notebook_file = notebook_file
+        # Convert absolute notebook path to relative path
+        if notebook_file[0] == '/':
```
karthikv2k: This will break in non-posix systems. use https://docs.python.org/3/library/os.path.html#os.path.isabs ?

rpasricha: Done

PR #175, opened 2019-04-03
Default Kubeflow backends to Job

### LINKED ISSUE #174: KubeflowBackend should not automatically choose TfJob
https://github.com/kubeflow/fairing/blob/master/fairing/backends/backends.py#L53

r2d4:
Fixes #174

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/175)


r2d4:
Job will work for any framework, but TfJob will only work for Tensorflow jobs.

Issue #176, opened 2019-04-03
 fairing fails to create an endpoint as arguments of get_service_external_endpoint() (name and namespace) are missing


zabbasi:
PredictionEndpoint fails with the following error 

```
TypeError                                 Traceback (most recent call last)
<ipython-input-10-8e16d12ce458> in <module>
      3 endpoint = PredictionEndpoint(HousingServe, BASE_IMAGE, input_files=['trained_ames_model.dat', "requirements.txt"],
      4                               docker_registry=DOCKER_REGISTRY, backend=GKEBackend())
----> 5 endpoint.create()

~/projects/fairingExamples2/fairing/venv/lib/python3.7/site-packages/fairing/ml_tasks/tasks.py in create(self)
     64         self._build()
     65         deployer = self._backend.get_serving_deployer(self.model_class.__name__)
---> 66         self.url = deployer.deploy(self.pod_spec)
     67         logger.warning("Prediction endpoint: {}".format(self.url))
     68 

~/projects/fairingExamples2/fairing/venv/lib/python3.7/site-packages/fairing/deployers/serving/serving.py in deploy(self, pod_spec)
     42 
     43         logger.warn("Endpoint {} launched.".format(self.deployment.metadata.name))
---> 44         url = self.backend.get_service_external_endpoint()
     45         return url
     46 

TypeError: get_service_external_endpoint() missing 2 required positional arguments: 'name' and 'namespace'
```

This happens when running examples/notebooks/prediction/xgboost-high-level-apis.ipynb from a local notebook.

PR #177, opened 2019-04-03
call get_service_external_endpoint correctly

r2d4:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/177)


r2d4:
Fixes https://github.com/kubeflow/fairing/issues/176

abhi-g:
We should figure out how unittests might help catch these.

PR #180, opened 2019-04-04
Adding necessary pod spec mutators to cluster builder in the high level api

karthikv2k:
Summary of changes:
1. Moved the guess builder logic to backend API. This let's us to customize builders based on the environment. For example, here we are mounting GCP service account creds to the cluster builder when user is in GCP. Once we have #145, we can make cloud build the default builder for GCP managed backend. #173 
1. Added delete method to serving deployer and high level api. This is required for adding an end2end test. 
1. Added an e2e test for examples/predict/xgboost-high-level-apis.
1. Had to put in a alternative to PredictionEndpoint.predict() in the xgboost example due to #179 but in our cloud build environment I was getting connection failure error to the prediction endpoint although i can predict against that endpoint from my local machine. It may be due to #126 as adding some time delay worked but did go through with it as it might create random failures . For now I added exception handling to skip this step from the e2e test.

Regarding tests:
The fix for #173 is for code path that gets invoked while in a K8s environment. Although our integration test environment have access to K8s cluster, the tests themselves don't currently run on K8s. Also this fix is specifically for notebook pods created by notebook controller. Therefore there is a non trivial effort to get the integration test based on the notebook controller. In the absence of automated testing, @abhi-g and I tested the fix manually on the latest KF master and it works.

There are few TODOs in this PR. To keep the PR small enough and get the P0 bug fixed ASAP, I made those things TODOs.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/180)

PR #181, opened 2019-04-04
Fix prediction

### LINKED ISSUE #179: Directly calling predict function in prediction endpoint is failing
```
(train_X, train_y), (test_X, test_y) = read_input('ames_dataset/train.csv')
endpoint.predict(test_X)
```
This fails as it tries to pass a numpy array directly to requests object. @abhi-g and I found it during our manual e2e testing.

r2d4:
fixes #179

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/181)

### DIFF
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -245,7 +245,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "(train_X, train_y), (test_X, test_y) = read_input([\"ames_dataset/train.csv\"])\n",
+    "(train_X, train_y), (test_X, test_y) = read_input(\"ames_dataset/train.csv\")\n",
     "endpoint.predict_nparray(test_X)\n"
    ]
   },
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -245,7 +245,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "(train_X, train_y), (test_X, test_y) = read_input('ames_dataset/train.csv')\n",
+    "(train_X, train_y), (test_X, test_y) = read_input(['ames_dataset/train.csv'])\n",
     "endpoint.predict_nparray(test_X)\n"
    ]
   },
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -245,7 +245,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "(train_X, train_y), (test_X, test_y) = read_input(['ames_dataset/train.csv'])\n",
+    "(train_X, train_y), (test_X, test_y) = read_input([\"ames_dataset/train.csv\"])\n",
     "endpoint.predict_nparray(test_X)\n"
    ]
   },
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -245,32 +245,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import requests\n",
-    "import json\n",
-    "json_data = {\n",
-    "    \"data\":{\n",
-    "        \"tensor\":{\n",
-    "            \"shape\":[1,37],\n",
-    "            \"values\":[1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19,20,21,22,23,24,25,26,27,28,29,30,31,32,33,34,35,36,37]\n",
-    "        }\n",
-    "    }\n",
-    "}\n",
-    "try:\n",
-    "    res=requests.post(endpoint.url, data={\"json\":json.dumps(json_data)})\n",
-    "    res_json = json.loads(res.text)\n",
-    "    names = res_json['data']['names']\n",
-    "    values = res_json['data']['tensor']['values']\n",
-    "    print(res_json)\n",
-    "except:\n",
-    "    pass\n",
-    "    "
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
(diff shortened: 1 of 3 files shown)


r2d4:
rebase on #180

fairing/ml_tasks/tasks.py
```diff
@@ -64,8 +65,18 @@ def create(self):
         self.url = self._deployer.deploy(self.pod_spec)
         logger.warning("Prediction endpoint: {}".format(self.url))
 
-    def predict(self, data):
-        r = requests.post(self.url, data=data)
+    def predict_nparray(self, data, names=["t:0", "t:1"]):
```
karthikv2k: Seldon expects the request body to be [SeldonMessage](https://github.com/SeldonIO/seldon-core/blob/36e96c162bfaa5c1fb5a3286d8c02dba8e393071/util/loadtester/scripts/proto/prediction.proto) proto. The names field represents the column/feature names.  Therefore we shouldn't default to any value here other than empty array or none. names here is passed to predict function as feature_names. However there is no check on len(names)==data.shape[1] in their code.
Ref for future reference to improve client impl:
https://github.com/SeldonIO/seldon-core/blob/7ad5eee98c2724c8ca27e9877304ad8c27a6ed84/python/seldon_core/seldon_methods.py#L38 
https://github.com/SeldonIO/seldon-core/blob/7ad5eee98c2724c8ca27e9877304ad8c27a6ed84/python/seldon_core/utils.py#L363
https://github.com/SeldonIO/seldon-core/blob/bb9ce7fd7aece435100ceb720bc9a5f421de786a/python/seldon_core/user_model.py#L133

fairing/ml_tasks/tasks.py
```diff
@@ -64,8 +65,18 @@ def create(self):
...
+        pdata={
+            "data": {
+                "names":names,
```
karthikv2k: names should be a 1D string array, so doing some check here would be useful.

fairing/ml_tasks/tasks.py
```diff
@@ -64,8 +65,18 @@ def create(self):
...
+                "tensor": {
+                    "shape": np.asarray(data.shape).tolist(),
+                    "values": data.tolist(),
```
karthikv2k: adding if data is a numerical array should be useful.

fairing/ml_tasks/tasks.py
r2d4: done

fairing/ml_tasks/tasks.py
r2d4: I don't think its expected that we do type validation everywhere

karthikv2k: The issue here is that if there is an error, it will be in the serving container logs which is hard to get to. From customer UX perspective, doing a small validation here goes a long way in avoiding a laborious debugging.

r2d4: we print out the server error, which should tell the user the same info

r2d4:
tests passed locally

fairing/ml_tasks/tasks.py
karthikv2k: The error I am getting if the names filed is something like this. I think this is fine for now but definitely can be improved with client side validation
```
'{"status":{"code":-1,"info":"Invalid JSON: Failed to parse names field: expected string or bytes-like object.","reason":"MICROSERVICE_BAD_DATA","status":1}}\n'
```

examples/prediction/xgboost-high-level-apis.ipynb
```diff
@@ -245,32 +245,8 @@
...
-    "    "
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "## Delete endpoint"
+    "(train_X, train_y), (test_X, test_y) = read_input([\"ames_dataset/train.csv\"])\n",
+    "endpoint.predict_nparray(test_X)\n"
```
abhi-g: questions for @karthikv2k : for the high level api, will Numpy arrays be the usual mode of data vars? Is this compatible with pandas?

karthikv2k: as an MVP numpy is fine. DS have familiarity with numpy. Definitely doing pandas support will be great and we can do it in 0.6 release.

karthikv2k: read_input(["ames_dataset/train.csv"]) -> read_input("ames_dataset/train.csv")?

fairing/ml_tasks/tasks.py
```diff
@@ -64,8 +65,18 @@ def create(self):
...
         logger.warning("Prediction endpoint: {}".format(self.url))
 
-    def predict(self, data):
-        r = requests.post(self.url, data=data)
+    def predict_nparray(self, data, feature_names=None):
+        pdata={
+            "data": {
+                "names":feature_names,
+                "tensor": {
+                    "shape": np.asarray(data.shape).tolist(),
```
karthikv2k: import numpy as np?

fairing/ml_tasks/tasks.py
```diff
@@ -64,8 +65,18 @@ def create(self):
...
+                    "values": data.tolist(),
```
karthikv2k: data.flatten().tolist()? SeldonMessage proto needs the values to be 1D array https://github.com/SeldonIO/seldon-core/blob/36e96c162bfaa5c1fb5a3286d8c02dba8e393071/util/loadtester/scripts/proto/prediction.proto#L33

r2d4: done

fairing/ml_tasks/tasks.py
r2d4: done

examples/prediction/xgboost-high-level-apis.ipynb
r2d4: done

abhi-g:
This worked for me :D I installed fairing from your branch @r2d4 and I was able to run through the high level api notebook from training to prediction!

PR #187, opened 2019-04-05
Test improvements and minor fixes for storage client auth

### LINKED ISSUE #183: Add numpy to requirements.txt
prediction client uses numpy but it is not captured in the requirements. In the test the numpy was installed as a side effect of our configuration so tests are not failing for this.

karthikv2k:
Summary:
1. A lot of tests were using constant message that might lead to false positives where a cached output makes the test pass. So I made the test messages to be UUIDs to make tests robust. Additionally, added some delays to print test messages to avoid test flakiness in case of very short lived jobs.
2. GCP Storage client was expecting a service account key and this was breaking in environments where application default credentials are used like GCE instances. So I made it to use underlying auth lib to figure out credential instead of requiring json key.
3. Fixes #183 
4. Updated python version used for tests to 3.7. Seldon only supports 3.6+ so it is better to update the version.
5. Changed base image used in tests to plain python images. This increases isolation between tests. Currently we use images that are built from previous tests runs that is causing unwanted dependencies.
6. Fixed a bug in the tests for docker and cluster builder where append builder was used instead of docker and cluster builder.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/187)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,4 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+.DS_Store
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,4 +1,4 @@
-FROM python:3.5
+FROM python:3.7
 
 WORKDIR /opt/fairing
 ARG CLOUD_SDK_VERSION="236.0.0"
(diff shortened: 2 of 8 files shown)


abhi-g:
/lgtm


We tested manually for Kubeflow notebooks and JaaS and local.

Issue #206, opened 2019-04-10
Don't expose services by default; Don't use Service Type LoadBalancer by default


jlewi:
It looks like by default when you deploy with fairing you create a service of type loadbalancer
https://github.com/kubeflow/fairing/blob/642530567d0255f88b0f83b0a40b3325c303bfa2/fairing/deployers/serving/serving.py#L79

This exposes the model on a public endpoint by default.

In general we want to avoid making services publicly available. Especially for fairing since the target audience (datascientists) may not know the security implications of doing so.

By default we should use ClusterIP.

To make the service accessible from outside the cluster we should add an appropriate reverse proxy route (e.g. via Ambassador or ISTIO). This way it will be secured via whatever authentication/authorization method the user has setup.

karthikv2k:
#153

r2d4:
Using a service mesh for routing would preclude fairing from deploying on vanilla Kubernetes clusters then. Is that ok?

jlewi:
@karthikv2k What is the plan for 0.6?

karthikv2k:
We don't have cycles to fix it in 0.6.

jinchihe:
This should be enhanced in #376 , closing the ticket.

Issue #200, opened 2019-04-10
Add auth to prediction to prediction


jlewi:
How about attaching authorization credentials to prediction requests?
e.g here:
https://github.com/kubeflow/fairing/blob/642530567d0255f88b0f83b0a40b3325c303bfa2/fairing/ml_tasks/tasks.py#L69

A use case would be you are developing your models locally e.g. in an IDE but deploying on your cluster and you want to send your prediction requests.

Kubeflow should be exposed through a reachable ingress that is secured (e.g. with IAP or Basic Auth). So the client library should obtain a credential (e.g. a JWT) and attach it to the request.

The pipelines SDK has an auth module
https://github.com/kubeflow/pipelines/blob/master/sdk/python/kfp/_auth.py

And there is an open issue to support IAP kubeflow/pipelines#1104

PR #207, opened 2019-04-11
Add logging, set working directory and service type.

jlewi:
* For AppendBuilder set working directory to /app so that files are on
  the python path

* Set default to ClusterIP for the service.
  Related to #206

  There's some ongoing discussion about the right long term solution but
  I think its better to lock things down in the short term.

* Add more logging.

* Print out the full in cluster URL of the service.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/207)

### DIFF
--- a/fairing/backends/backends.py
+++ b/fairing/backends/backends.py
@@ -1,5 +1,4 @@
-import abc
-import logging
+import abc 
 import six
 
 import fairing
@@ -35,32 +34,29 @@ class KubernetesBackend(BackendInterface):
 
     def __init__(self, namespace=None):
         self._namespace = namespace
-
+    
     def get_builder(self, preprocessor, base_image, registry, needs_deps_installation=True, pod_spec_mutators=None):
         if fairing.utils.is_running_in_k8s():
-            logging.info("Using ClusterBuilder")
             return ClusterBuilder(preprocessor=preprocessor,
                                   base_image=base_image,
                                   registry=registry,
                                   pod_spec_mutators=pod_spec_mutators,
                                   namespace=self._namespace)
         elif ml_tasks_utils.is_docker_daemon_exists():
-            logging.info("Using Docker builder")
             return DockerBuilder(preprocessor=preprocessor,
                                  base_image=base_image,
                                  registry=registry)
         elif not needs_deps_installation:
-            logging.info("Using AppendBuilder")
             return AppendBuilder(preprocessor=preprocessor,
                                  base_image=base_image,
                                  registry=registry)
         else:
             # TODO (karthikv2k): Add more info on how to reolve this issue
             raise RuntimeError("Not able to guess the right builder for this job!")
-
+    
     def get_training_deployer(self):
         return Job(self._namespace)
-
+    
     def get_serving_deployer(self, model_class):
         return Serving(model_class, namespace=self._namespace)
 
@@ -68,7 +64,7 @@ class GKEBackend(KubernetesBackend):
 
     def __init__(self, namespace=None):
         super(GKEBackend, self).__init__(namespace)
-
+    
     def get_builder(self, preprocessor, base_image, registry, needs_deps_installation=True, pod_spec_mutators=None):
         pod_spec_mutators = pod_spec_mutators or []
         pod_spec_mutators.append(gcp.add_gcp_credentials_if_exists)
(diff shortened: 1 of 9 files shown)


fairing/builders/base_builder.py
```diff
@@ -51,7 +51,10 @@ def generate_pod_spec(self):
                 env=[client.V1EnvVar(
                     name='FAIRING_RUNTIME',
                     value='1',
-                )]
+                )],
+                # Set the directory where the python files are built.
+                # TODO(jlewi): Would it be better to set PYTHONPATH?
+                working_dir="/app",
```
r2d4: this should be `self.preprocessor.path_prefix`

fairing/deployers/serving/serving.py
```diff
@@ -41,7 +41,10 @@ def deploy(self, pod_spec):
         self.deployment = apps_v1.create_namespaced_deployment(self.namespace, self.deployment_spec)
         self.service = v1_api.create_namespaced_service(self.namespace, self.service_spec)
 
-        logger.warn("Endpoint {} launched.".format(self.deployment.metadata.name))
+        logger.warn("Deployment {} launched.".format(self.deployment.metadata.name))
+        url = "http://{0}.{1}.svc.cluster.local".format(self.service.metadata.name,
```
r2d4: This won't always be the service endpoint. the suffix `cluster.local` is configurable in Kubernetes

fairing/deployers/serving/serving.py
```diff
@@ -76,13 +80,13 @@ def generate_service_spec(self):
                     name="serving",
                     port=5000
                 )],
-                type="LoadBalancer",
+                type="ClusterIP",
```
r2d4: ClusterIP is only routable from inside the cluster. In Kubeflow notebooks, thats ok. For local + other notebooks, this won't work.

jlewi: For outside the cluster can we reference the ingress for the cluster which should be secured?

I filed some issues for this:
#208 - Add ISTIO routing rules
#200 - Add auth to prediction requests 

So with GCP for example we'd print the full URL
https://${APP}.endpoints.${PROJECT}.cloud.goog

And then attach a JWT for IAP to the prediction request.

r2d4: I see how it can be made to point to the IAP endpoint on GCP, but thats not the change proposed. And how would this work for generic use cases?

Changing to clusterIP without making those other changes would break all users

jlewi: I made it an argument with a default of ClusterIP. Does that work?

Folks depending on LoadBalancer will need to update the code. Do we have examples that would need to be updated. 

We could leave the default as type LoadBalancer but I'd prefer not to be "open" by default. Especially since the target audience (data scientists) is unlikely to understand the risks of doing so.

ClusterIp will work for in cluster traffic and for port-forwarding. 

For external traffic I was thinking users would do one of the following
  1. Create a suitable ingress pointing at the Kubeflow ISTIO gateway (akin to IAP)
  1. Chose to expose the service either via type LoadBalancer or via Ingress

If folks want to do #2 I think its fine I just think they should have to explicitly choose to do so.

fairing/builders/base_builder.py
jlewi: Done

fairing/deployers/serving/serving.py
jlewi: Good to know. Is there a simple way to get the in cluster DNS name correctly? 

We could also return the empty string if you think that's better.

fairing/backends/backends.py
```diff
@@ -34,37 +35,40 @@ class KubernetesBackend(BackendInterface):
...
                                   registry=registry,
                                   pod_spec_mutators=pod_spec_mutators,
                                   namespace=self._namespace)
         elif ml_tasks_utils.is_docker_daemon_exists():
+            logging.info("Using Docker builder")
             return DockerBuilder(preprocessor=preprocessor,
                                  base_image=base_image,
                                  registry=registry)
         elif not needs_deps_installation:
+            logging.info("Using AppendBuilder")
```
r2d4: Let's put the log messages in the builders themselves so it doesn't have to be replicated everywhere we instantiate them.

fairing/preprocessors/base.py
```diff
@@ -84,9 +84,13 @@ def context_map(self):
         return c_map
 
     def context_tar_gz(self, output_file=constants.DEFAULT_CONTEXT_FILENAME):
+        logging.info("Creating docker context: %s", output_file)
         self.input_files = self.preprocess()
+        logging.info("Adding files to context: %s", self.input_files)
```
r2d4: `context_map` files are the full set of files added to the context

fairing/backends/backends.py
jlewi: Done.

fairing/preprocessors/base.py
jlewi: Done.

jlewi:
Thanks @r2d4

PR #214, opened 2019-04-15
update documentation on xgboost example notebook

r2d4:
cc @abhi-g

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/214)

### DIFF
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


joeliedtke:
It would be great to understand a bit more about this change. When I read the notebook, the important detail to me is that users can train on Kubeflow on GKE and Cloud ML Engine, then deploy to Kubeflow on GKE.

Would the instructions be different if the user was training on Kubeflow on a different Kubernetes backend? If that is the case, then I would recommend calling this out as a GCP specific example.

It may be worth calling it out as a GCP example in either case because it also includes code to train on CMLE. What do you think?

PR #215, opened 2019-04-16
Fix a typo in reference to service_type to fix postsubmit tests

jlewi:
* Switch default service type back to LoadBalancer because the notebook
  test currently depends on that and there's no easy way to override it
  in the notebook.

* I didn't actually run the post-submit tests because I'm not sure how.

Related to #206

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/215)

### DIFF
--- a/fairing/deployers/serving/serving.py
+++ b/fairing/deployers/serving/serving.py
@@ -18,8 +18,12 @@ class Serving(Job):
 
     """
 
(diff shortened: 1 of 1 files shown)


karthikv2k:
I will run the post commit tests and see if things work.

PR #218, opened 2019-04-22
Making tests run parallel

### LINKED ISSUE #213: Run tests in parallel using pytest-xdist

karthikv2k:
Fixes #213 
This reduces test run time from 20-21 mins to 6-7 mins in cloud build including the cloud build start up time. 

Changes summary:
1. Made the preprocessor output file random instead of a constant. This make is possible to run multiple fairing jobs in parallel in an instance. This is useful for users who are launching multiple jobs in different notebooks in the same instance and also for tests.
2. Bumping up the python version in travis to match post commit test python version
3. Printing top 10 slowest test for improving test run times
4. Moved requests lib to the top of requirements.txt to resolve some deps mismatch issue
5. Using machine type with 8 vCPUs for post commit test. Test machine type with 32 vCPUs (only other choice) but didn't improve the test times considerably.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/218)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 # Python versions to test
 python:
-    - "3.5"
+    - "3.7"
 # Command to install dependencies
 install:
     - python setup.py install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial # required for Python >= 3.7
 language: python
 # Python versions to test
 python:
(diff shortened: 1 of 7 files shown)


abhi-g:
It looks the presubmit failure is due to a non-configured test workflow in the repo.

the change otherwise lgtm

r2d4:
@abhi-g should be fixed by https://github.com/kubeflow/fairing/pull/220

Issue #1, opened 2018-09-03
Setup Prow for this repository


jlewi:
See:
https://github.com/kubeflow/community/blob/master/repository-setup.md

jlewi:
@r2d4 Has this been fixed?

jlewi:
@karthikv2k @r2d4  I see #221 is pending; does that mean prow has been setup and we can close this issue?

PR #221, opened 2019-04-22
Add prow tests

r2d4:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/221)


r2d4:
Fixes #1

karthikv2k:
@r2d4 is there a doc on how this setup works? if not can you write a one page summary about how it works?

tests/e2e.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
karthikv2k: Can you add a doc string here?

tests/unit.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
```
karthikv2k: Can you add a doc string here?

tests/e2e.sh
```diff
@@ -0,0 +1,6 @@
...
+
+set -e
+
+gcloud auth activate-service-account --key-file=/secret/gcp-credentials/key.json
+# gcloud builds submit --config cloudbuild.yaml .
```
karthikv2k: where are we running the integration tests? I only see unit tests are bing run in unit.sh

.travis.yml
```diff
@@ -1,13 +0,0 @@
-dist: xenial # required for Python >= 3.7
```
karthikv2k: Let's keep the travis setup for some time till we see no issues with prow presubmit, probably for a couple of weeks.

tests/e2e.sh
r2d4: I disabled the integration tests because I wasn't able to get them running successfully. The infrastructure does run tests successfully, but theres probably a difference in environment/bug in the tests that hangs.

r2d4:
For reference: Here's how the integration tests are set up in this PR

The kubernetes/kubernetes prow installation is set up to trigger a single presubmit test on this repository for pull requests.

That prow job is configured to use the base image that kubeflow uses across its tests and the "run_e2e.py" script that prepares the argo workflow that is contained in this PR.

The argo workflow is currently split into two steps - unit and integration tests. The unit tests run directly on the kubeflow testing infrastructure GKE cluster, while the integration tests trigger the same cloud build job we were previously running.

The cloud build job runs on a new kubeflow v0.5 cluster that runs in the kubeflow-ci project.

abhi-g:
closing and cloning from this branch to take over the setup. Will create a new PR.

PR #226, opened 2019-04-25
Added tests for converted notebook preprocessor

karthikv2k:
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/226)

### DIFF
--- a/tests/unit/preprocessors/test_converted_notebook_preprocessor.py
+++ b/tests/unit/preprocessors/test_converted_notebook_preprocessor.py
@@ -23,7 +23,6 @@ def test_get_command():
     command = preprocessor.get_command()
     conv_notebook_context_path = posixpath.join(DEFAULT_DEST_PREFIX, CONVERTED_NOTEBOOK_PATH)
     expected_command = 'python {}'.format(conv_notebook_context_path)
-    print(command)
     assert command == expected_command.split()
 
 def test_context_tar_gz():
--- a/tests/unit/preprocessors/test_converted_notebook_preprocessor.py
+++ b/tests/unit/preprocessors/test_converted_notebook_preprocessor.py
@@ -1,0 +1,36 @@
+import pytest
+import tarfile
+import os
+import posixpath
+from pathlib import Path
+
+import fairing
+from fairing.preprocessors.converted_notebook import ConvertNotebookPreprocessor
+from fairing.constants.constants import DEFAULT_DEST_PREFIX
+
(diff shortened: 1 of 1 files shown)


tests/unit/preprocessors/test_converted_notebook_preprocessor.py
```diff
@@ -0,0 +1,36 @@
...
+    files = preprocessor.preprocess()
+    assert Path(CONVERTED_NOTEBOOK_PATH) in files
+
+def test_get_command():
+    preprocessor = ConvertNotebookPreprocessor(notebook_file=NOTEBOOK_PATH)
+    preprocessor.preprocess()
+    command = preprocessor.get_command()
+    conv_notebook_context_path = posixpath.join(DEFAULT_DEST_PREFIX, CONVERTED_NOTEBOOK_PATH)
+    expected_command = 'python {}'.format(conv_notebook_context_path)
+    print(command)
```
abhi-g: Can we add a prefix to this print value such as "Command: "

tests/unit/preprocessors/test_converted_notebook_preprocessor.py
```diff
@@ -0,0 +1,36 @@
...
+NOTEBOOK_PATH = os.path.relpath(
+    os.path.join(os.path.dirname(__file__), 'test_notebook.ipynb'))
+CONVERTED_NOTEBOOK_PATH = NOTEBOOK_PATH.replace(".ipynb",".py")
+
+def test_preprocess():
+    preprocessor = ConvertNotebookPreprocessor(notebook_file=NOTEBOOK_PATH)
+    files = preprocessor.preprocess()
+    assert Path(CONVERTED_NOTEBOOK_PATH) in files
+
+def test_get_command():
```
abhi-g: Does pytest execute the tests in order? i.e. when executing this one, does it guarantee that previous tests passed or were at least ran? that way if test_preprocess fails, we can ignore errors from this test.

karthikv2k: No. It shouldn't by design. I think what you are asking for is a smoke test but these are unit tests so we wont incur time running all the tests so no gating mechanism is necessary. However, we can do something for integration tests where unit or some other tests failures would not run the remaining tests.

tests/unit/preprocessors/test_converted_notebook_preprocessor.py
karthikv2k: it was a debug statement so removed it.

Issue #236, opened 2019-05-07
integration tests using tf-job is hanging


karthikv2k:
this blocking post commit tests to timeout

karthikv2k:
The job is getting launched and log streaming is getting stuck.

karthikv2k:
the issue seems to be change in label format in tf-controller https://github.com/kubeflow/tf-operator/commit/6f748507e0cafe787cb358b802176a53b6b11040

Issue #194, opened 2019-04-10
Define __version__ in fairing package


jlewi:
Should we define "fairing.__version__" so people can easily tell what version of fairing they have installed?

This might be useful for customer support; e.g. asking what version a user is using.

jlewi:
@karthikv2k Can we close this issue?

PR #238, opened 2019-05-08
 Adding fairing specific user agent to GCP calls

### LINKED ISSUE #232: Self identify requests coming from Fairing SDK
We need to add some instrumentation to identify requests/job submissions to KF and other backends to enable debugging and monitoring on the backend side. An example would be setting the user-agent in the http requests to "fairing<version>". This enable the ops team to find if there is a specific version of Fairing causing any issues and also to deprecate older versions.

karthikv2k:
Fixes #232, #236, #194

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/238)


Dockerfile.test
```diff
@@ -5,7 +5,7 @@ ARG CLOUD_SDK_VERSION="236.0.0"
 
 # Install pinned version of gcloud
 RUN mkdir -p /builder && \
-	wget -qO- https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${CLOUD_SDK_VERSION}-linux-x86_64.tar.gz | tar zxv -C /builder && \
+	wget -qO- https://dl.google.com/dl/cloudsdk/channels/rapid/downloads/google-cloud-sdk-${CLOUD_SDK_VERSION}-linux-x86_64.tar.gz | tar zx -C /builder && \
```
karthikv2k: reducing log verbosity

fairing/deployers/tfjob/tfjob.py
```diff
@@ -67,6 +67,6 @@ def get_logs(self):
         labels = {
             'tf-replica-index': '0',
             'tf-replica-type': 'worker',
-            'tf_job_name=': name
+            'tf-job-name': name
```
karthikv2k: due to https://github.com/kubeflow/tf-operator/commit/6f748507e0cafe787cb358b802176a53b6b11040

abhi-g: thanks for catching this

fairing/http_utils.py
```diff
@@ -0,0 +1,28 @@
...
+       
+       request_orig = http.request
+       user_agent = constants.DEFAULT_USER_AGENT.format(VERSION=fairing.__version__)
+       # Reference: https://github.com/googleapis/google-api-python-client/blob/master/googleapiclient/http.py
+       # The closure that will replace 'httplib2.Http.request'.
+       def new_request(*args, **kwargs):
+              """Modify the request headers to add the user-agent."""
+              if args and len(args)>=4:
+                     headers = args[3] or {}
+                     headers['user-agent'] = user_agent
```
abhi-g: In case the request already has a User-Agent, should this be overwriting or appending?https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent 
Possibly appending with a space delimiter might be better?

fairing/http_utils.py
```diff
@@ -0,0 +1,28 @@
...
+                     args[3] = headers
+              else:
+                     headers = kwargs.get('headers') or {}
+                     headers['user-agent'] = user_agent
```
abhi-g: same as above

tests/unit/deployers/gcp/test_gcp.py
```diff
@@ -1,8 +1,12 @@
 import pytest
+import re
 import fairing
 
+import httplib2
 from kubernetes import client
 from fairing.deployers.gcp.gcp import GCPJob
+from unittest.mock import patch
+from googleapiclient import discovery
```
abhi-g: unused imports?

tests/unit/deployers/gcp/test_gcpserving.py
```diff
@@ -1,6 +1,7 @@
 """Tests for GCPServing Deployer."""
 
 import pytest
+import re
```
abhi-g: unused import

fairing/http_utils.py
karthikv2k: i did that first then looking at how attribution is done, it seems to be done based on regex match so I thought it might get duplicate attribution with google python client. On the second thought that might be ok.

tests/unit/deployers/gcp/test_gcp.py
karthikv2k: Ack. Initially the tests were here and then I moved them out but forgot to delete the imports

Issue #242, opened 2019-05-10
Add support for Lightgbm


rexshang:
Support to train model in parallel is needed for Lightgbm.  This will help shorten the time required to train a bigger model.

PR #259, opened 2019-05-22
Added distributed training support for LightGBM

karthikv2k:
Summary:
1. Added distribution training support for LightGBM #242 
1. Minor change to requirements.txt in XGBoost example to get its test passing
1. Added integration test for distributed training and prediction tasks

Next tasks for distributed support:
1. Ability to partition and copy training data to worker nodes.
1. Documentation
1. Benchmark distributed training performance

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/259)

### DIFF
--- a/examples/lightgbm/distributed-training.ipynb
+++ b/examples/lightgbm/distributed-training.ipynb
@@ -1,0 +1,188 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## A simple regression training using LightGBM through Fairing"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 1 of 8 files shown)


abhi-g:
@karthikv2k I am seriously jetlagged and exhausted, so this is not a detailed code review. 

At a high level the code to support distribution looks correct. But I feep like majority of the utils code looks like can fit an existing k8s pattern such as an operator? If so, can you please add a todo comment to rewrite the code with that pattern. It should be possible to implement in Python as well.


In the meantime i don't want to block your progress and you can go ahead and merge this.

/Lgtm

karthikv2k:
Thanks @abhi-g . I created #260 to capture your comment. Will talk with the team to find a good way to refractor the code.

PR #263, opened 2019-05-29
Implement AWSBackend and S3Context

takmatsu:
This patch enables kubeflow-fairing to run in AWS environment(EKS, S3, ECR...) like this.

```
import os
import fairing

DOCKER_REGISTRY = '{account_id}.dkr.ecr.{region}.amazonaws.com'
PY_VERSION = ".".join([str(x) for x in sys.version_info[0:3]])
BASE_IMAGE = 'python:{}'.format(PY_VERSION)
```
```
from fairing import TrainJob
from fairing.backends import KubeflowAWSBackend
train_job = TrainJob(HousingServe, BASE_IMAGE, input_files=['ames_dataset/train.csv', "requirements.txt"],
                     docker_registry=DOCKER_REGISTRY, backend=KubeflowAWSBackend())
train_job.submit()
```

Fixes #282 

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/263)

PR #266, opened 2019-05-31
Support hostname as a serving endpoint

takmatsu:
Some LB-controller implementation (like ELB) returns not ip but hostname.
This patch adds support of hostname as serving endpoint url.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/266)

### DIFF
--- a/fairing/kubernetes/manager.py
+++ b/fairing/kubernetes/manager.py
@@ -76,7 +76,7 @@ class KubeManager(object):
                              event['object'])
                 ing = svc.status.load_balancer.ingress
                 if ing is not None and len(ing) > 0:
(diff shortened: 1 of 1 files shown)


karthikv2k:
@takmatsu thanks for your contribution. Are you planning to add more features related to running Fairing on AWS? if so we should probably setup an integration test environment for Fairing + KF on AWS. Also please create issues for any major work you are planning to do in the future so that we can suggest and review the design before you put in the effort.

/lgtm
/approve

takmatsu:
@karthikv2k Thank you for your review. currently all my work is PR #263

PR #267, opened 2019-05-31
LightGBM - Copies weight file if it exists

karthikv2k:
Summary:
1. LightGBM uses .weight file automatically if it exists so to keep the behavior same, .weight file copied from GCS if it exists. https://lightgbm.readthedocs.io/en/latest/Parameters.html#weight-data 
1. Refactored storage utils and created an abstraction to support storage systems other than GCS
1. Added unit tests
1. Created and uploaded .weight file for the integration test

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/267)

### DIFF
--- a/fairing/cloud/storage.py
+++ b/fairing/cloud/storage.py
@@ -1,0 +1,74 @@
+import six
+import abc
+from google.cloud import storage
+from urllib.parse import urlparse
+
+def lookup_storage_class(url):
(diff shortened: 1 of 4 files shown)


abhi-g:
Thanks @karthikv2k 

/lgtm
/approve

PR #272, opened 2019-06-07
add name to container

hamedhsn:
right now when we use job deployer the container named randomly. This will fix the name. 
pls suggest if my chosen name is not appropriate.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/272)


hamedhsn:
@karthikv2k can you approve pls?

karthikv2k:
Thanks for your contribution.
/lgtm
/approve

PR #278, opened 2019-06-13
remove useless set_model

jinchihe:

The `set_model` function has been removed in PR #114 , but the main.py files are not updated. Seems only need to call fairing.run, no need set_model. The PR is to remove the set_model calling from main.py. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/278)


karthikv2k:
Thanks for your contribution.
/lgtm
/approve

PR #279, opened 2019-06-13
Adding different entry point options to Train Job API

### LINKED ISSUE #203: Why doesn't PredictionEndpoint/Task interface support the append builder?
It looks like the PredictionEndpoint interface doesn't support the append builder.

It doesn't look like there's a way to pass to get_builder `needs_deps_installation=False` to force it to use the append builder rather than the ClusterBuilder (Kaniko)

https://github.com/kubeflow/fairing/blob/642530567d0255f88b0f83b0a40b3325c303bfa2/fairing/ml_tasks/tasks.py#L36

/cc @karthikv2k

karthikv2k:
1. Currently TrainJob API only supports Python function/class. This PR adds support for executing python files and full notebook execution.
2. Exposes pod_spec_mutator argument in TrainJob API that lets user to customize pod spec like CPUs and mount volumes
3. Improved how we pick the builder in TrainJob API. Now it uses AppendBuilder when no python package installation is required. Also makes sure right gcp secret is available before executing cluster builder.
4. Improved error messages
5. Added an example notebook - examples/train_job_api/main.ipynb

Fixes #203 #166 

NA

**Release note**:

```release-note
TrainJob API supports python file and full notebook execution.
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/279)

### DIFF
--- a/fairing/backends/backends.py
+++ b/fairing/backends/backends.py
@@ -11,6 +11,8 @@ from fairing.deployers.job.job import Job
 from fairing.deployers.serving.serving import Serving
 from fairing.cloud import gcp
 import fairing.ml_tasks.utils as ml_tasks_utils
+from fairing.constants import constants
+from fairing.kubernetes.manager import KubeManager
 
 @six.add_metaclass(abc.ABCMeta)
 class BackendInterface(object):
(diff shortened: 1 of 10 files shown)


fairing/backends/backends.py
```diff
@@ -109,7 +146,11 @@ def get_builder(self, preprocessor, base_image, registry, needs_deps_installatio
         pod_spec_mutators.append(gcp.add_gcp_credentials_if_exists)
         # TODO (karthikv2k): Add cloud build as the deafult
         # once https://github.com/kubeflow/fairing/issues/145 is fixed
-        if fairing.utils.is_running_in_k8s():
+        if not needs_deps_installation:
```
abhi-g: This logic for deciding which Builder to use would be good to factor into a separate function that returns an Enum or equivalent in python. that would make it easier to keep in sync at multiple invocations.

abhi-g:
I like the template you have going here for the description of the issue! we should discuss this in one of the community meetings.

karthikv2k:
> I like the template you have going here for the description of the issue! we should discuss this in one of the community meetings.

credit goes to @gyliu513 #270

fairing/backends/backends.py
karthikv2k: I made the changes but let me do that as a separate PR to keep the changes to minimum.

PR #281, opened 2019-06-17
Add ability to specify context_source for k8s backends

takmatsu:
Adding ability to specify context_source for kubernetes-related backends.
Kaniko supports not only GCS context, but S3 and GitHub.
For future implementation, we need the ability to change that.

Related #282 

**Release note**:

```release-note
NONE
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/281)

### DIFF
--- a/fairing/backends/backends.py
+++ b/fairing/backends/backends.py
@@ -46,9 +46,9 @@ class BackendInterface(object):
 
 class KubernetesBackend(BackendInterface):
 
-    def __init__(self, namespace=None, context_source=None):
+    def __init__(self, namespace=None, build_context_source=None):
         self._namespace = namespace
-        self._context_source = context_source
+        self._build_context_source = build_context_source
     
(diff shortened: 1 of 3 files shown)


karthikv2k:
I think we need to use better descriptive name instead of context_source in the backends api. Something like container_context_source? Also currently if the context_source is none then we fall back to GCS which should be modified to through an error and make the context_source a required arg in cluster builder class.

in GKEBackend and KubeflowGKEBackend, this should be defaulted to GCS context source and in KubernetesBackend and KubeflowBackend this should be made a required arg. 

Thoughts?

takmatsu:
@karthikv2k I totally agree with you.
How is `build_context_source`?

karthikv2k:
Thanks for your contribution @takmatsu . 

/lgtm
/approve

PR #285, opened 2019-06-18
update examples to add set_preprocessor

jinchihe:

When I warming up the Kubeflow Fairing, and try some examples for on-prem cluster, I found the defined function/class is not built in the image, so while create pod from the image, the training is not run (some of training function run locally). That's not right. Checked details, I found the need to `set_preprocessor` and point the `function_obj` in the `set_preprocessor`. The RP is to update the example to enhance this. Thanks.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/285)

### DIFF
--- a/examples/distributed-training/main.py
+++ b/examples/distributed-training/main.py
@@ -31,8 +31,6 @@ import logging
 import tensorflow as tf
 from tensorflow.examples.tutorials.mnist import input_data
 
-import fairing
-
 
 MAX_STEPS = 1000
 LEARNING_RATE = 0.001
@@ -230,6 +228,12 @@ class TensorflowModel(object):
 
 
 if __name__ == '__main__':
-    fairing.config.set_builder(name='docker', registry='gcr.io/mrick-gcp', base_image='tensorflow/tensorflow')
-    fairing.config.set_deployer(name='tfjob', namespace='default', worker_count=1, ps_count=1)
-    fairing.config.run()
+    if os.getenv('FAIRING_RUNTIME', None) is None:
+        import fairing
+        fairing.config.set_preprocessor('python', input_files=[__file__])
+        fairing.config.set_builder(name='docker', registry='gcr.io/mrick-gcp', base_image='tensorflow/tensorflow')
+        fairing.config.set_deployer(name='tfjob', namespace='default', worker_count=1, ps_count=1)
+        fairing.config.run()
+    else:
+        remote_train = TensorflowModel()
+        remote_train.train()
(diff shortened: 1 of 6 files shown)


jinchihe:
/assign @karthikv2k 

Could you please review? Thanks a lot!

karthikv2k:
Function preprocessor is good for notebook use case but not so for working with python files. The main restriction for function preprocessor is that the python version should be same for both local python and the python in the docker. For executing code in python files from cli the recommended way is to use base preprocessor or trainjob api with entrypoint pointing to a python file. If you are executing the current python file then you should also use an if statement to separate job launching from invoking the train method. something like this:

```
import os

if os.getenv('FAIRING_RUNTIME', None) is not None:
    fairing.config.set_preprocessor('base', input_files=[__file__])
    fairing.config.set_builder('append', base_image='tensorflow/tensorflow:1.13.1-py3')
    fairing.config.run()
else:
    train()
```

or
```
import os

if os.getenv('FAIRING_RUNTIME', None) is not None:
    job = TrainJob(__file__, backend=KubeflowGKEBackend())
    job.submit()
else:
    train()
```

jinchihe:
@karthikv2k Great thanks for the pointing out. Yes, that's good for base the `BasePreProcessor`. 

And that's good to use an if statement to separate job launching from invoking the train method, but seems your example above is inverse :-) In fact in out local (build) env, no variables `FAIRING_RUNTIME`, and we can import `fairing` and build/push image; and in the docker contrainer, has `FAIRING_RUNTIME` so just run the `python /app/main.py`. So I think the if statement should be like this:

```
if os.getenv('FAIRING_RUNTIME', None) is None:
    import fairing 
    fairing.config.set_preprocessor('python', input_files=[__file__])
    fairing.config.set_builder('append', base_image='tensorflow/tensorflow:1.13.1-py3')
    fairing.config.run()
else:
    train()
```

I'm going to update the code changes, you know, if current code, the file will not pacakged into docker image at all. Thanks again.

jinchihe:
Hi @karthikv2k,

Updated code that includes below:
1. Using `FAIRING_RUNTIME` to separate job launching from invoking the train method.
2. Using `fairing.config.set_preprocessor('python', input_files=[__file__])` so that the `main.py` can be pacakged in the docker image.
3. import fairing only in local fairing building env (in if statement), the docker container should no (and no need) the fairing installed.
4. Removed useless variables `MODEL_DIR` in the examples.

Thanks a lot!

karthikv2k:
Thanks for catching the bug in my example and for the contribution.

/lgtm
/approve

PR #291, opened 2019-06-19
Check python version for function preprocessor

### LINKED ISSUE #288: Check python version during execution for function preprocessor
/kind feature

Function preprocessor needs the python version to be the same for both local and remote python. If version mismatch, users get not a useful error message, something like unknown opcode. We should check the version mismatch before executing the payload in the function shim and output a useful message.

jinchihe:

Check python version during execution for function preprocessor.

In the patch we just compare the Python major and minor version, and igore micro version, for examples, 3.6.5 will mismatch with 3.5.2, but match with 3.6.8. 

After the patch, if the major and minor version are same, then continue. Otherwise, will exit and print beflow error message from pod:
```
[root@jinchi1 simple]# kubectl logs fairing-job-55rzc-bz9fh -n default
  File "test.py", line 24, in <module>
    compare_version(args.python_version)
  File "test.py", line 18, in compare_version
    + local_python_version + ' in the local environment.')
RuntimeError: The Python version 3.6 mismatches with Python 3.5 in the local environment.
```

Fixes #288

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/291)

### DIFF
--- a/fairing/functions/function_shim.py
+++ b/fairing/functions/function_shim.py
@@ -1,5 +1,6 @@
 import argparse
 import cloudpickle
+import sys
 import types
 from enum import Enum
 import logging
@@ -29,6 +30,14 @@ def get_execution_obj_type(obj):
 
     return ObjectType.NOT_SUPPORTED
 
+# compare the Python major and minor version for local and remote python
+def compare_version(local_python_version):
+    remote_python_version = ".".join([str(x) for x in sys.version_info[0:3]])
+    if local_python_version != remote_python_version:
+        raise RuntimeError('The Python version ' + remote_python_version + ' mismatches with Python '
+                        + local_python_version + ' in the local environment.')
+
+
 def call(serialized_fn_file):
     with open(serialized_fn_file, 'rb') as f:
         obj = cloudpickle.load(f)
@@ -45,5 +54,7 @@ def call(serialized_fn_file):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Shim for calling a serialized function')
     parser.add_argument('--serialized_fn_file', action="store")
+    parser.add_argument('--python_version', action="store")
     args = parser.parse_args()
+    compare_version(args.python_version)
     call(args.serialized_fn_file)
--- a/tests/integration/common/test_py_version_checking.py
+++ b/tests/integration/common/test_py_version_checking.py
@@ -1,0 +1,48 @@
+import pytest
+import fairing
+import sys
+import io
+import tempfile
+import random
+import time
+import uuid
+
+GCS_PROJECT_ID = fairing.cloud.gcp.guess_project_name()
+DOCKER_REGISTRY = 'gcr.io/{}'.format(GCS_PROJECT_ID)
+
+# Dummy training function to be submitted
+def train_fn(msg):
+    for _ in range(30):
+        time.sleep(0.1)
+        print(msg)
+
+# Update module to work with function preprocessor
+train_fn.__module__ = '__main__'
+
+def run_submission_with_function_preprocessor(capsys, base_image, is_match, namespace="default"):
+    fairing.config.set_builder("append", base_image=base_image, registry=DOCKER_REGISTRY)
+    fairing.config.set_deployer("job", namespace=namespace)
+
+    test_result = str(uuid.uuid4())
+    remote_train = fairing.config.fn(lambda : train_fn(test_result))
+    remote_train()
+    captured = capsys.readouterr()
+    if is_match:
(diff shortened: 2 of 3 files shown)


jinchihe:
/assign @karthikv2k 

Could you please take a look? Thanks a lot！

fairing/functions/function_shim.py
```diff
@@ -29,6 +30,14 @@ def get_execution_obj_type(obj):
 
     return ObjectType.NOT_SUPPORTED
 
+# compare the Python major and minor version for local and remote python
+def compare_version(local_python_version):
+    remote_python_version = ".".join([str(x) for x in sys.version_info[0:2]])
+    if local_python_version != remote_python_version:
+        logging.error("ERROR: The Python version mismatches with the local environment.")
+        sys.exit(1)
```
karthikv2k: throwing an exception makes sense here. sys.exit usually done in the entrypoint of the program like the main() not usually in a function.

fairing/functions/function_shim.py
```diff
@@ -45,5 +54,7 @@ def call(serialized_fn_file):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Shim for calling a serialized function')
     parser.add_argument('--serialized_fn_file', action="store")
+    parser.add_argument('--python_version', action="store")
     args = parser.parse_args()
+    compare_version(args.python_version)
```
karthikv2k: we can do a try catch here but i think leaving as it is would give the same intended behavior. 
```
$ python -c "import sdsfd"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'sdsfd'
$ echo $?
1
```

fairing/functions/function_shim.py
jinchihe: Addressed your comments. Thanks.

fairing/functions/function_shim.py
jinchihe: keep the traceback without catch, so user will see as below:

```
  File "test.py", line 24, in <module>
    compare_version(args.python_version)
  File "test.py", line 18, in compare_version
    + local_python_version + ' in the local environment.')
RuntimeError: The Python version 3.6 mismatches with Python 3.5 in the local environment.
```

jinchihe:
> Can you please add integration tests that checks for 3 cases (major version mismatch, minor version mismatch, and no mismatch)?

@karthikv2k Thanks. In fact I added two cases (mismatching and matching), since personally I think that's enough, two reasons:
1.  From our document, Fairing supports running for python 3.x. And during my testing, if used python 2.7, there are some problems, such as will get Traceback for the "No module named enum" since the enum is not installed in the python 2.x by default, **before python version checking.**
2. From coding logic, check the 'major.minor', not separate checking for major version and minor version.

Thanks again.

tests/integration/common/test_py_version_checking.py
```diff
@@ -0,0 +1,48 @@
...
+
+def run_submission_with_function_preprocessor(capsys, base_image, is_match, namespace="default"):
+    fairing.config.set_builder("append", base_image=base_image, registry=DOCKER_REGISTRY)
+    fairing.config.set_deployer("job", namespace=namespace)
+
+    test_result = str(uuid.uuid4())
+    remote_train = fairing.config.fn(lambda : train_fn(test_result))
+    remote_train()
+    captured = capsys.readouterr()
+    if is_match == 'Y':
```
karthikv2k: make is_match boolean?

karthikv2k:
My mistake in my prev comment. I wanted to check if the build version (sys.version_info[2]) need to be the same for the fn preprocessor to work. If I remember right, all three components of the version should match in order for the fn preprocessor to work.

tests/integration/common/test_py_version_checking.py
jinchihe: Done, thanks

jinchihe:
> My mistake in my prev comment. I wanted to check if the build version (sys.version_info[2]) need to be the same for the fn preprocessor to work. If I remember right, all three components of the version should match in order for the fn preprocessor to work.

Oh...  I tested the 3.6.5 (local) and 3.6.8 (remote), so I just checked the major and minor, igore micro version (take 3.6.5 as example: major=3, minor=6, micro=5). so if you think we also need to check the micro version, I will update to handle that, but seems too strict, because some basic image is tensorflow or something, that's hard to find right micro version for the basic image except user build that from right python :-)

jinchihe:
@karthikv2k Addressed your comments to check full python version, thanks.

PR #292, opened 2019-06-21
Warn users who are using function preprocessor and not in notebook

### LINKED ISSUE #289: Warn users who are using function preprocessor for executing logic in python files
/kind feature

Function preprocessor is good for notebook use case but not so for working with python files. The main restriction for function preprocessor is that the python version should be same for both local python and the python in the docker. For executing code in python files the recommended way is to use base preprocessor or trainjob api with entrypoint pointing to a python file. 

Adding a warning if we see the function preprocessor in an environment other than notebook or ipython shell would be useful.

jinchihe:
Adding a warning if user is using the function preprocessor in non-netebook environment, as below:

```
The FunctionPreProcessor is optimized for using in a notebook or IPython environment. For it to work, the python version should be same for both local python and the python in  the docker. Please look at alternatives like BasePreprocessor or FullNotebookPreprocessor.
```

Fixes #289

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/292)

### DIFF
--- a/fairing/preprocessors/function.py
+++ b/fairing/preprocessors/function.py
@@ -1,6 +1,7 @@
 import cloudpickle
 import fairing
 import glob
+import logging
 import os
 import sys
 import types
@@ -11,6 +12,9 @@ from enum import Enum
 from fairing.constants import constants
 from .base import BasePreProcessor
 from fairing.functions.function_shim import get_execution_obj_type, ObjectType
+from fairing.notebook import notebook_util
+
+logger = logging.getLogger(__name__)
 
 FUNCTION_SHIM = 'function_shim.py'
 SERIALIZED_FN_FILE = 'pickled_fn.p'
@@ -36,6 +40,11 @@ class FunctionPreProcessor(BasePreProcessor):
             output_map=output_map,
             path_prefix=path_prefix,
             input_files=input_files)
+
+        if not notebook_util.is_in_notebook():
+            logger.warning("The FunctionPreProcessor is optimized for using in a notebook or IPython environment. "
+                           "For it to work, the python version should be same for both local python and the python in "
+                           "the docker. Please look at alternatives like BasePreprocessor or FullNotebookPreprocessor.")
 
         if get_execution_obj_type(function_obj) ==  ObjectType.NOT_SUPPORTED:
             raise RuntimeError("Object must of type function or a class")


jinchihe:
/assign @karthikv2k
Could you please help to review the simple enhancement, thanks.

fairing/preprocessors/function.py
```diff
@@ -36,6 +40,9 @@ def __init__(self,
             path_prefix=path_prefix,
             input_files=input_files)
 
+        if notebook_util.is_in_notebook():
```
karthikv2k: if notebook_util.is_in_notebook(): -> if not notebook_util.is_in_notebook():?

fairing/preprocessors/function.py
```diff
@@ -36,6 +40,9 @@ def __init__(self,
...
+            logger.warning("The FunctionPreProcessor is typically used for notebook environment.")
```
karthikv2k: Adding little more detail would be helpful. "The Function PreProcessor is optimized for using in a notebook or IPython environment. For it to work, the python version should be same for both local python and the python in the docker. Please look at alternatives like BasePreprocessor or FullNotebookPreprocessor."

Being more verbose with error/warn messages is always a good thing.

fairing/preprocessors/function.py
jinchihe: Oh...apology... In my test env, I added 'not', forgot that for the pushing...

fairing/preprocessors/function.py
jinchihe: Great warning message!

karthikv2k:
Can you rebase and try running the test?

jinchihe:
Checked intergration test logs, seems caused by env problem while running `test_cluster_builder`?  Try one more time.
```
W
Step #2: _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
Step #2:
Step #2: self = <kubernetes.config.kube_config.ConfigNode object at 0x7f70f61f4e48>
Step #2: key = 'current-context'
Step #2:
Step #2:     def safe_get(self, key):
Step #2:         if (isinstance(self.value, list) and isinstance(key, int) or
Step #2: >               key in self.value):
Step #2: E               TypeError: argument of type 'NoneType' is not iterable
Step #2:
Step #2: /usr/local/lib/python3.7/site-packages/kubernetes/config/kube_config.py:459: TypeError
```

/retest

PR #299, opened 2019-06-24
support kfserving in fairing

### LINKED ISSUE #290: Support KFServing to deploy model service
/kind feature

Suggest Fairing supporting kubeflow [KFServing](https://github.com/kubeflow/kfserving), KFserving has support few usual ML frameworks (such as tensorflow, scikitLearn, pyTorch), and that can be used to deploy a service quickly and easily from trained model and suppport rollout deployments gently. And kubeflow KFServing aims to solve 80% of model serving use cases, so think that's a good choise to deploy ML serving.

@karthikv2k @r2d4 any thoughts for this? Thanks.

jinchihe:
Support KFServing in Fairing.

Fixes #290 

1. The intergration tests for the kfserving are not added, since the automation platform should not have the kfserving installation. But I tested most of cases in the on-prem cluster, and we can add later once the kfserving is installed for test platform.

2. I updated the `log()` function in the `fairing/kubernetes/manager.py` to satify the implements, since there are two container in the pod, we need to point out which container logs will be got.

**Release note**:

```release-note
Support KFServing to deploy model service
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/299)

### DIFF
--- a/examples/kfserving/main.py
+++ b/examples/kfserving/main.py
@@ -1,0 +1,21 @@
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import fairing
+
+if __name__ == '__main__':
+        fairing.config.set_deployer('kfserving', namespace='kubeflow', framework='tensorflow',
+                                   default_model_uri='gs://kfserving-samples/models/tensorflow/flowers',cleanup=True)
+        fairing.config.run()
--- a/fairing/constants/constants.py
+++ b/fairing/constants/constants.py
@@ -21,3 +21,13 @@ TF_JOB_PLURAL = "tfjobs"
 TF_JOB_VERSION = "v1beta2"
 TF_JOB_DEFAULT_NAME = 'fairing-tfjob-'
 TF_JOB_DEPLOYER_TYPE = 'tfjob'
+
+# KFServing constants
+KFSERVING_GROUP = "serving.kubeflow.org"
+KFSERVING_KIND = "KFService"
+KFSERVING_PLURAL = "kfservices"
+KFSERVING_VERSION = "v1alpha1"
+KFSERVING_DEFAULT_NAME = 'fairing-kfserving-'
+KFSERVING_DEPLOYER_TYPE = 'kfservice'
+KFSERVING_CONTAINER_NAME = 'user-container'
+
--- a/fairing/kubernetes/manager.py
+++ b/fairing/kubernetes/manager.py
@@ -37,6 +37,31 @@ class KubeManager(object):
         api_instance = client.AppsV1Api()
         return api_instance.create_namespaced_deployment(namespace, deployment)
 
+    def create_kfserving(self, namespace, kfservice):
+        """Create the provided KFServing in the specified namespace"""
+        api_instance = client.CustomObjectsApi()
+        try:
+            return api_instance.create_namespaced_custom_object(
+                constants.KFSERVING_GROUP,
+                constants.KFSERVING_VERSION,
+                namespace,
+                constants.KFSERVING_PLURAL,
+                kfservice)
+        except client.rest.ApiException:
+            raise RuntimeError("Failed to create KFService. Perhaps the CRD KFServing version "
+                               "{} is not installed?".format(constants.KFSERVING_VERSION))
+
+    def delete_kfserving(self, name, namespace):
+        """Delete the provided KFServing in the specified namespace"""
+        api_instance = client.CustomObjectsApi()
+        return api_instance.delete_namespaced_custom_object(
+            constants.KFSERVING_GROUP,
+            constants.KFSERVING_VERSION,
+            namespace,
+            constants.KFSERVING_PLURAL,
+            name,
+            client.V1DeleteOptions())
+
     def delete_job(self, name, namespace):
         """Delete the specified job"""
         api_instance = client.BatchV1Api()
@@ -83,7 +108,7 @@ class KubeManager(object):
         except client.rest.ApiException as e:
             logger.error("error getting status for {} {}".format(name, str(e)))
 
-    def log(self, name, namespace, selectors=None):
+    def log(self, name, namespace, selectors=None, container='', follow=True):
         label_selector_str = ', '.join("{}={}".format(k, v) for (k, v) in selectors.items())
         v1 = client.CoreV1Api()
         # Retry to allow starting of pod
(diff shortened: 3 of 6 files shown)


jinchihe:
/assign @karthikv2k 

Could you please review? Thanks a lot!

fairing/constants/constants.py
```diff
@@ -13,3 +13,12 @@
...
 DEFAULT_USER_AGENT = 'kubeflow-fairing/{VERSION}'
+
+# KFServing constants
+KFSERVING_GROUP = "serving.kubeflow.org"
+KFSERVING_KIND = "KFService"
+KFSERVING_PLURAL = "kfservices"
+KFSERVING_VERSION = "v1alpha1"
+KFSERVING_DEFAULT_NAME = 'fairing-kfserving-'
+KFSERVING_DEPLOYER_TYPE = 'kfservice'
+KFSERVING_CONTAINER_NAME = 'user-container'
```
karthikv2k: nit: add a new line?

examples/kfserving/main.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+import fairing
+
+if __name__ == '__main__':
+        fairing.config.set_builder(name='docker', registry='<Your-Docker-Registry>',
+                                   base_image='python:3.6.5')
+        fairing.config.set_deployer('kfserving', namespace='kubeflow', framework='tensorflow',
+                                   default_model_uri='gs://kfserving-samples/models/tensorflow/flowers',cleanup=True)
+        fairing.config.run()
```
karthikv2k: nit: add a new line?

examples/kfserving/main.py
```diff
@@ -0,0 +1,23 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+import fairing
+
+if __name__ == '__main__':
+        fairing.config.set_builder(name='docker', registry='<Your-Docker-Registry>',
```
karthikv2k: do you need a builder here?

fairing/kubernetes/manager.py
```diff
@@ -87,7 +107,7 @@ def get_service_external_endpoint(self, name, namespace, selectors=None):
         except client.rest.ApiException as e:
             logger.error("error getting status for {} {}".format(name, str(e)))
 
-    def log(self, name, namespace, selectors=None):
+    def log(self, name, namespace, selectors=None, container='', follow=True):
```
karthikv2k: container=None?

fairing/constants/constants.py
jinchihe: @karthikv2k Thanks. Updated. I'm introducing the pylint tests for fairing, once dine, those problem will be checked, :-)

examples/kfserving/main.py
jinchihe: Done.

fairing/kubernetes/manager.py
jinchihe: I think that should be "container=''", I tested "container=None", if the container is not defined (you know the parameter is new, and some old callings are not set this), the API prints error say no this container, but "" is OK. Thanks.

examples/kfserving/main.py
jinchihe: I set the builder since I met the connecting exceptions with gcr.io because of GOOGLE_APPLICATION_CREDENTIALS no configuration. If set the my registry, will no the problem, so I just suggest keep this since other users may met the problem. Agree? @karthikv2k Thanks.

jinchihe:
@karthikv2k Thanks for you great comments. Yes we should definitely add the intergration testing for kfserving, but you know the ci cluster has no kfserving installation. I just created the ticket #302 and assigned to me. Firstly we need to install kfserving on the testing cluster but seems not easy since that depends on isito and knative. But will going on to check or work for this later. Thanks.

karthikv2k:
Can you add a unit test mocking necessary K8s client and comparing the produced CRD to the expected CRD for kfserving?

karthikv2k:
>I just created the ticket #302 and assigned to me.

Sounds good. We can add the integration test later.

karthikv2k:
please refer to existing unit tests for working with mock clients. e.g. https://github.com/kubeflow/fairing/blob/master/tests/unit/cloud/test_gcp_creds.py

jinchihe:
> Can you add a unit test mocking necessary K8s client and comparing the produced CRD to the expected CRD for kfserving?

@karthikv2k Added some unit test cases for kfserving. Thanks.

jinchihe:
> how does fairing think about debugging? There are many cases where this might fail to deploy (dependencies not installed, no cpu/gpu available, etc). 

Thanks @ellis-bigelow, I just updated code to rasie useful message for user if the kfserving created failed (such as the version is not installed). Once the kfservice created, we have `get_log()` function that shows logs from pod so user can get more information, and I think user can check kfservice status if the kfservice is not started normally.

> Is Fairing intended to be only the golden path?

I'm not sure I understand exactly what the meaning for `golden path`. Fairing supports kfservice getting model only not from `golden path`, but also from local trained. You know kserving is going to support from minio and pvc, once tested done, Failing also support from there.

May @karthikv2k has more accurate and professional comments for this :-)

jinchihe:
@karthikv2k Any more comments for this? Thanks!

abhi-g:
Karthik is on vacation, please wait until he is back next week.

jinchihe:
> Karthik is on vacation, please wait until he is back next week.

Oh... Thanks for your reminder, haha. Wish @karthikv2k a happly vacation :-)

examples/kfserving/main.py
karthikv2k: We should write this example by directly calling the Serving class. There is no need to preprocess or build steps here.

tests/unit/deployers/kfserving.py
```diff
@@ -0,0 +1,44 @@
...
+from fairing.deployers.kfserving.kfserving import KFServing
+
+# Test kfserving function with default_model_uri.
+def test_kfserving_default_model_spec():
+    kfsvc_name = 'test_kfservice'
+
+    with patch('fairing.deployers.kfserving.kfserving.KFServing.deploy', return_value=kfsvc_name):
+        kfsvc = KFServing(framework='tensorflow',
+                          default_model_uri='gs://kfserving-samples/models/tensorflow/flowers')
+        assert kfsvc.deploy("test") == kfsvc_name
```
karthikv2k: we need to test more than just the service name. Testing the output of generate_kfservice is useful here.

karthikv2k:
> @karthikv2k , how does fairing think about debugging? There are many cases where this might fail to deploy (dependencies not installed, no cpu/gpu available, etc). Is Fairing intended to be only the golden path?

Fairing only provides only basic debugging like during the resource creation. If there are no gpus in the cluster or some other dependencies are not installed then the user have to use the GKE tools to debug. 

> Is Fairing intended to be only the golden path?
yes. debugging your code within a pod is partially support by streaming the logs but for full debugging users have to use kubectl or gke dash.

examples/kfserving/main.py
jinchihe: Removed builder here. Thanks.

tests/unit/deployers/kfserving.py
jinchihe: Updated the unit test file to check more from generate_kfservice outputs. Thanks.

jinchihe:
@karthikv2k Addressed your comments, updated the `tests/unit/deployers/kfserving.py`, and removed set_builder from example files. Thanks a lot!

PR #311, opened 2019-07-08
Add a option to let user cleanup tfjob

### LINKED ISSUE #239: tf-job pods are not cleaned after job finishes

jinchihe:
Add a option to let user cleanup tfjob after tfjobs finishes.

Fixes #239

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/311)

### DIFF
--- a/tests/integration/common/test_kubeflow_training.py
+++ b/tests/integration/common/test_kubeflow_training.py
@@ -8,10 +8,13 @@ import time
 import uuid
 
 from google.cloud import storage
+from kubernetes import client
+
 from fairing import TrainJob
 from fairing.backends import KubernetesBackend, KubeflowBackend
 from fairing.backends import KubeflowGKEBackend, GKEBackend, GCPManagedBackend
 from fairing.builders.cluster import gcs_context
+from fairing.constants import constants
 
 GCS_PROJECT_ID = fairing.cloud.gcp.guess_project_name()
 DOCKER_REGISTRY = 'gcr.io/{}'.format(GCS_PROJECT_ID)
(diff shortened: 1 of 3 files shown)


karthikv2k:
Can you add a unit/integration test?

jinchihe:
@karthikv2k I added one integration tests for clean up `TFJob`, in the case I just checked there is "Cleaning Up" message in the `caplog`, I think that's enough since the following step just calling python client APIs.

By the way, I also added one   integration cases to test clean up `job`. Thank you.

tests/integration/common/test_kubeflow_training.py
```diff
@@ -36,22 +36,32 @@ def run_submission_with_function_preprocessor(capsys, deployer="job", builder="a
...
-    fairing.config.set_deployer(deployer, namespace=namespace)
+    fairing.config.set_deployer(deployer, namespace=namespace, cleanup=cleanup)
 
     expected_result = str(uuid.uuid4())
     remote_train = fairing.config.fn(lambda : train_fn(expected_result))
     remote_train()
     captured = capsys.readouterr()
     assert expected_result in captured.out
+    if cleanup:
+        assert "Cleaning up" in caplog.text
```
karthikv2k: it is better to check if the resources are deleted using k8s api instead of just checking the logs. You can imagine the case where we might print "cleaning up" but the resources may not be cleaned up.

jinchihe: @karthikv2k Thanks. Yes, That's better. I'm lazy, haha... Rollbacked last changes, and called k8s python api to check if the tfjob object is removed if set cleanup.

jinchihe:
By the way, from current implement, seems if have labels defined, the labels is not set to TFJob metadata but set to Cheif/Works/Ps metadata, the labels should also be set in the TFJob metadata. The RP also enhance this.

jinchihe:
Seems the testing environment has problem now.

jinchihe:
Rebased to resolve the conflicts, thanks.

PR #320, opened 2019-07-18
Avoid converting to different path if running preprocessor multiple times

### LINKED ISSUE #319: ConvertNotebookPreprocessorWithFire doesn't properly handle case where converted notebook exists
/kind bug

[A clear and concise description of what the bug is.]

* Run through the [example](https://github.com/kubeflow/examples/blob/master/xgboost_synthetic/build-train-deploy.ipynb)

* The first time you run it; the file `build-train-deploy.py` is created
* The second time you run it we trigger this [code path](https://github.com/kubeflow/fairing/blob/f43a4a3675e43cb736b690a07f22490b54523867/fairing/preprocessors/converted_notebook.py#L126)
* Since the file `build-train-deploy.py` exists we end up trying to create a temporary file
* There is a bug in this code and "converted_notebook" ends up getting set to the empty file which is problematic

We should fix this

* We should avoid avoid triggering a different code path if you run multiple times; that leads to hard to catch bugs
* Instead lets just overwrite the file 
* If we want to guard against accidental rewriting we should add a flag to not overwrite it if it already exists

abcdefgs0324:
- Avoid converting to different path if we run preprocessors multiple times
- Change path to the absolute path to fix path error like converting notebook to ".py"

Fixes #319 

/assign @jlewi 
/assign @zhenghuiwang

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/320)

### DIFF
--- a/fairing/preprocessors/converted_notebook.py
+++ b/fairing/preprocessors/converted_notebook.py
@@ -74,7 +74,7 @@ class ConvertNotebookPreprocessor(BasePreProcessor):
         contents, _ = exporter.from_filename(self.notebook_file)
         converted_notebook = Path(self.notebook_file).with_suffix('.py')
         if converted_notebook.exists() and not self.overwrite:
-            raise Exception('Default path {} has existed but overwrite flag is False'.format(converted_notebook))
+            raise Exception('Default path {} exists but overwrite flag is False'.format(converted_notebook))
         with open(converted_notebook, 'w') as f:
             f.write(contents)
         self.executable = converted_notebook
@@ -125,7 +125,7 @@ class ConvertNotebookPreprocessorWithFire(ConvertNotebookPreprocessor):
         contents = "\n".join(lines)
         converted_notebook = Path(self.notebook_file).with_suffix('.py')
         if converted_notebook.exists() and not self.overwrite:
-            raise Exception('Default path {} has existed but overwrite flag is False'.format(converted_notebook))
+            raise Exception('Default path {} exists but overwrite flag is False'.format(converted_notebook))
         with open(converted_notebook, 'w') as f:
             f.write(contents)
             f.write("\n")
--- a/fairing/preprocessors/converted_notebook.py
+++ b/fairing/preprocessors/converted_notebook.py
@@ -65,8 +65,12 @@ class ConvertNotebookPreprocessor(BasePreProcessor):
 
         self.notebook_file = notebook_file
         self.notebook_preprocessor = notebook_preprocessor
+        self.not_overwrite = False
 
     def preprocess(self):
+        if self.not_overwrite:
+            return [self.executable]
+        self.not_overwrite = True
         exporter = nbconvert.PythonExporter()
         exporter.register_preprocessor(self.notebook_preprocessor, enabled=True)
         contents, _ = exporter.from_filename(self.notebook_file)
@@ -101,8 +105,14 @@ class ConvertNotebookPreprocessorWithFire(ConvertNotebookPreprocessor):
             output_map=output_map)
 
         self.class_name = class_name
+        self.not_overwrite = False
 
     def preprocess(self):
+        if self.not_overwrite:
+            results =  [self.executable]
+            results.extend(self.input_files)
+            return results
+        self.not_overwrite = True
         exporter = nbconvert.PythonExporter()
         exporter.register_preprocessor(self.notebook_preprocessor, enabled=True)
         processed, _ = exporter.from_filename(self.notebook_file)
--- a/fairing/preprocessors/converted_notebook.py
+++ b/fairing/preprocessors/converted_notebook.py
@@ -52,7 +52,8 @@ class ConvertNotebookPreprocessor(BasePreProcessor):
                  executable=None,
                  command=["python"],
                  path_prefix=constants.DEFAULT_DEST_PREFIX,
-                 output_map=None):
+                 output_map=None,
+                 overwrite=True):
 
         super().__init__(
             executable=executable,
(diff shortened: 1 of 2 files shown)


fairing/preprocessors/converted_notebook.py
```diff
@@ -101,8 +105,14 @@ def __init__(self,
             output_map=output_map)
 
         self.class_name = class_name
+        self.not_overwrite = False
 
     def preprocess(self):
+        if self.not_overwrite:
+            results =  [self.executable]
```
jinchihe: minio:  extra space here after "="

abcdefgs0324: Fixed.

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +65,20 @@ def __init__(self,
 
         self.notebook_file = notebook_file
         self.notebook_preprocessor = notebook_preprocessor
+        self.not_overwrite = False
```
zhenghuiwang: Can you make `not_overwrite` in `__init__` with a default value? or it is intended to be a private property (in this case `_not_overwrite`)?

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +65,20 @@ def __init__(self,
...
 
     def preprocess(self):
+        if self.not_overwrite:
```
zhenghuiwang: Can you add a test case for the new code here? Maybe just passing in various flag in the tests here:

https://github.com/kubeflow/fairing/blob/master/tests/unit/preprocessors/test_full_notebook_preprocessor.py#L12

fairing/preprocessors/converted_notebook.py
abcdefgs0324: I think it should be a private property. I will update it.

fairing/preprocessors/converted_notebook.py
abcdefgs0324: Sure, I will add a test case for the flag.

tests/unit/preprocessors/test_converted_notebook_preprocessor.py
```diff
@@ -21,6 +21,14 @@ def test_preprocess():
     os.remove(converted_notebook_path)
     assert Path(converted_notebook_path) in files
 
+def test_not_overwrite_flag():
```
zhenghuiwang: Nit: name it "test_not_overwrite_file_for_multiple_runs"?

abcdefgs0324: Fixed, thanks.

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +65,20 @@ def __init__(self,
 
         self.notebook_file = notebook_file
         self.notebook_preprocessor = notebook_preprocessor
+        self._not_overwrite = False
```
jlewi: Can we pass this in via the constructor so its easy to override?

jlewi: Lets default it to true.

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +65,20 @@ def __init__(self,
...
 
     def preprocess(self):
+        if self._not_overwrite:
```
jlewi: If overwrite is False and the file exists lets throw an exception rather than just assuming the existing file is up to date.

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +65,20 @@ def __init__(self,
...
+            return [self.executable]
+        self._not_overwrite = True
         exporter = nbconvert.PythonExporter()
         exporter.register_preprocessor(self.notebook_preprocessor, enabled=True)
         contents, _ = exporter.from_filename(self.notebook_file)
         converted_notebook = Path(self.notebook_file).with_suffix('.py')
         if converted_notebook.exists():
-            dir_name = Path(self.notebook_file).parent
+            dir_name = Path(self.notebook_file).absolute().parent
```
jlewi: We still have two different code paths  i.e. the if block will not be executed if the file doesn't already exist.

I think the desired behavior is

1. Check if file exists
1. If file exists and overwrite=False; raise exception
1. Otherwise overwrite the file

In either case we should always use the same path for the file we write to.

fairing/preprocessors/converted_notebook.py
abcdefgs0324: `self._not_overwrite` is initialized to be `False`.
After the first pass of `preprocess()`, `self._not_overwrite` is set to `True` no matter what the output path is (e.g., notebook_file.py or a temporary file, which is stored as `self.executable`).
And if we call `preprocess()` for multiple times, we do not output the file again because of this check `if self._not_overwrite:`.

It's a way to convert notebook to python script even though the default path has existed and the desired behavior you mentioned below is a way to confirm the output path does not change.
How do you think which one is better?

fairing/preprocessors/converted_notebook.py
jlewi: Why is it a private property? If its a private property how will it get set and by whom?

abcdefgs0324: One preprocessor object only converts notebook to python script for one time. If the default path has existed, it will generate a temporary file.
`sefl._not_overwrite` checks whether it is the first time to call `preprocess()`. So, it only gets set when initialized.

fairing/preprocessors/converted_notebook.py
```diff
@@ -65,16 +66,15 @@ def __init__(self,
...
         exporter = nbconvert.PythonExporter()
         exporter.register_preprocessor(self.notebook_preprocessor, enabled=True)
         contents, _ = exporter.from_filename(self.notebook_file)
         converted_notebook = Path(self.notebook_file).with_suffix('.py')
-        if converted_notebook.exists():
-            dir_name = Path(self.notebook_file).parent
-            _, file_name = tempfile.mkstemp(suffix=".py", dir=dir_name)
-            converted_notebook = Path(file_name[file_name.find(str(dir_name)):])
+        if converted_notebook.exists() and not self.overwrite:
+            raise Exception('Default path {} has existed but overwrite flag is False'.format(converted_notebook))
```
jlewi: nit: "has existed" -> "exists"

fairing/preprocessors/converted_notebook.py
```diff
@@ -122,10 +124,8 @@ def preprocess(self):
 
         contents = "\n".join(lines)
         converted_notebook = Path(self.notebook_file).with_suffix('.py')
-        if converted_notebook.exists():
-            dir_name = Path(self.notebook_file).parent
-            _, file_name = tempfile.mkstemp(suffix=".py", dir=dir_name)
-            converted_notebook = Path(file_name[file_name.find(str(dir_name)):])
+        if converted_notebook.exists() and not self.overwrite:
+            raise Exception('Default path {} has existed but overwrite flag is False'.format(converted_notebook))
```
jlewi: nit: "has existed" -> "exists"

tests/unit/preprocessors/test_converted_notebook_preprocessor.py
```diff
@@ -21,6 +21,14 @@ def test_preprocess():
     os.remove(converted_notebook_path)
     assert Path(converted_notebook_path) in files
 
+def test_overwrite_file_for_multiple_runs():
```
jlewi: Thanks for adding a test.

fairing/preprocessors/converted_notebook.py
abcdefgs0324: Fixed

fairing/preprocessors/converted_notebook.py
abcdefgs0324: Fixed

jlewi:
Looks like only @abhi-g  or @karthikv2k can approve this.

PR #322, opened 2019-07-19
Azure backend support

vjrantal:

This PR contains implementation for `fairing.backends.KubeflowAzureBackend` which allows storing build context to Azure Storage and using Azure Container Registry as the Docker Registry.

1. An experimental CI is being worked on at https://dev.azure.com/kubeflow/kubeflow/_build with the intent of running Azure-specific integration tests there.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/322)


vjrantal:
Still working with some cleanup and tests, but wanted get feedback from the community to following questions:

1. The PR contains an E2E sample notebook https://github.com/vjrantal/fairing/blob/d7a82de6d34a8c58750a5ad10976756bf298169b/examples/prediction/xgboost-high-level-apis-azure.ipynb which uses the implemented `KubeflowAzureBackend`. However, the sample is very similar to the `xgboost-high-level-apis.ipynb` which does the same thing against the Google backend. I was wondering would it make sense to consolidate into a single sample? The notebook defaults could remain as they are now, but it could support parameters to use different backends similarly as I used in this test https://github.com/vjrantal/fairing/blob/d7a82de6d34a8c58750a5ad10976756bf298169b/tests/integration/azure/test_running_in_notebooks.py#L51. The value would be less code duplication and a single sample might also better highlight the benefit of the fairing SDK in terms of re-using same code on different environments.
2. The PR contains an E2E integration test at https://github.com/vjrantal/fairing/blob/d7a82de6d34a8c58750a5ad10976756bf298169b/tests/integration/azure/test_running_in_notebooks.py which I have tested to work on Azure. However, I believe the way the current test infrastructure works, that test would be tried to be run there as well, which would not work (the required Kubernetes Secrets wouldn't be setup there). I was wondering what would be a good approach for backend-specific integration tests?

fairing/kubernetes/manager.py
```diff
@@ -113,6 +113,7 @@ def get_service_external_endpoint(self, name, namespace, selectors=None):
             logger.error("error getting status for {} {}".format(name, str(e)))
 
     def log(self, name, namespace, selectors=None, container='', follow=True):
+        tail = ''
```
Jeffwan: clean up unused vars?

vjrantal: Thanks for the review @Jeffwan!

The line was added because that variable is referred to at line https://github.com/vjrantal/fairing/blob/azure-support/fairing/kubernetes/manager.py#L163 but assigned only within certain if-blocks. Without the added line, the error message I got was:

```
The local variable `tail` was referenced before assignment
```

The root cause for the error is the same as what one would see if executing this kind of Python code https://gist.github.com/vjrantal/17c57a1ff48b08a7a497b8061805996a.

If you have suggestions how to refactor the `log`-function in another way to get rid of this issue, I can amend my commit accordingly.

examples/prediction/xgboost-high-level-apis-azure.ipynb
```diff
@@ -0,0 +1,349 @@
+{
```
jinchihe: Is that better to update the current notebook file to add a cell for Azure backends? since there are some duplicated code and if each backend introduce one, that's not good for maintain.

jinchihe:
@vjrantal Please sign CLA as above mentioned. Seems the code has already done, if yes, please remove [WIP] from title for reviewing, thanks.

/ok-to-test

examples/prediction/xgboost-high-level-apis-azure.ipynb
vjrantal: @jinchihe This was actually my question at https://github.com/kubeflow/fairing/pull/322#issuecomment-513196543 and I would personally prefer a single notebook. This is also one reason I marked as WIP, but if the consensus is a single notebook, I am happy to change the implementation.

vjrantal:
@jinchihe Thanks for issuing the CI test. I am assuming the build will fail related to question 2. at https://github.com/kubeflow/fairing/pull/322#issuecomment-513196543.

The Azure integration tests will not work in the current test infrastructure, because the correct secrets are not available in the Kubernetes cluster to provision Azure resources.

I am thinking that perhaps short term, I should exclude Azure integration tests from this PR and perhaps separately open a discussion how to execute automated tests against multiple cloud environments?

jinchihe:
> @jinchihe Thanks for issuing the CI test. I am assuming the build will fail related to question 2. at [#322 (comment)](https://github.com/kubeflow/fairing/pull/322#issuecomment-513196543).
> 
> The Azure integration tests will not work in the current test infrastructure, because the correct secrets are not available in the Kubernetes cluster to provision Azure resources.
> 
> I am thinking that perhaps short term, I should exclude Azure integration tests from this PR and perhaps separately open a discussion how to execute automated tests against multiple cloud environments?

Personally I think we just should add some unit test cases to use `unittest.mock` for Azure backend resource. Here may help you: https://github.com/kubeflow/fairing/blob/bc0174f0483fa08320d8db773b377f8fe7017a30/tests/unit/deployers/kfserving.py#L11

vjrantal:
@jinchihe I have done mocking for Azure unit tests such as here https://github.com/vjrantal/fairing/blob/azure-support/tests/unit/cloud/test_azure.py. My concern above was specific only to the integration tests where the intent was not to mock, but to execute the code against real cloud environments.

magencio:
@googlebot I consent

cloudbuild.yaml
```diff
@@ -19,7 +19,7 @@ steps:
   entrypoint: 'bash'
   args: 
       - '-c'
-      - 'gcloud container clusters get-credentials kubeflow-ci-fairing --zone us-central1-a && pytest -n 8 -v --durations=10 --cov=fairing tests/'
+      - 'gcloud container clusters get-credentials kubeflow-ci-fairing --zone us-central1-a && pytest -n 8 -v --durations=10 --cov=fairing --ignore=tests/integration/azure/ tests/'
```
abhi-g: Is there going to a way to test the azure integrations in future?

examples/prediction/xgboost-high-level-apis-azure.ipynb
abhi-g: It makes sense to have at least have one single notebook that demonstrates all backends. Apart from that we may create notebooks highlighting specific backends if needed.

My only concern is that if any of the integration tests are exercising this notebook we need to ensure that the single notebook with all backends is able to execute periodically as part of the integration tests.

cloudbuild.yaml
vjrantal: We have an experimental CI setup here https://dev.azure.com/kubeflow/kubeflow that runs them. You can see an example of a success run here https://dev.azure.com/kubeflow/kubeflow/_build/results?buildId=38.

From my perspective, it would be great if these could be run in the official CI, but that would require setting up the right secrets in that environment and agreeing how to deal with the small cost associated with actually running things in Azure (cc: @aronchick).

abhi-g:
@vjrantal your 2nd point about integration tests with different cloud providers is worth discussing in slack. We may get good ideas from people who have thought about this. Try posting in #general first.

abhi-g:
@vjrantal please update if you need further input. Otherwise let's try to get this merged.

vjrantal:
@karthikv2k The integration test run failed to following error when trying to use the `GCPManagedBackend` in the notebook `xgboost-high-level-apis.ipynb`:

```
MaxRetryError: HTTPConnectionPool(host='35.238.254.1', port=5000): Max retries exceeded with url: /predict (Caused b
y NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f667c08c750>: Failed to establish a new connection: [Errno 110] Con
nection timed out'))
```

This is actually a new test that didn't exist before due to the way I parametrized the notebook. Before, the notebook only used `GCPManagedBackend` for the training part, but not for the deployment part.

What I wanted to check is that is it a supported scenario to use `GCPManagedBackend` for deployment?

vjrantal:
@abhi-g I have now refactored onto a single notebook. I think the parametrized approach does work and here https://dev.azure.com/kubeflow/kubeflow/_build/results?buildId=47 is a run that executes that notebook with the Azure backend.

The only negative effect is that there is now one cell in the notebook that is slightly more "complicated" than what it used to be due to having the conditional backend choice. Still, I would say that having notebook that highlights how little changes are required to switch to different backends is valuable.

vjrantal:
This time I see the following from the output:

```
Step #2: [gw2] [  8%] PASSED tests/integration/gcp/test_gcp_high_level_apis.py::test_job_submission_gcpmanaged
Step #2: tests/integration/gcp/test_running_in_notebooks.py::test_xgboost_highlevel_apis_gke
Step #2: [gw0] [  9%] PASSED tests/integration/common/test_high_level_apis.py::test_job_submission_kubernetesbackend
```

I wonder if that means that all tests in the file `test_running_in_notebooks.py` passed?

Overall, the e2e tests timed out, but I was not able to determine from the logs whether the changes in this PR caused the issue.

jinchihe:
> Overall, the e2e tests timed out, but I was not able to determine from the logs whether the changes in this PR caused the issue.

@vjrantal Recent days, the presubmit tests has problems, pls see #342 for details. We are going to resolve the problem now, but seems that's test env problem.

abhi-g:
@vjrantal please resolve merge conflicts and retry. The failures now should be real and relevant. We have the timeout issue fixed and various PRs have been able to merge.

vjrantal:
@abhi-g I can see this in the logs of the failed job:

```
Step #2: [gw5] [ 74%] PASSED tests/unit/preprocessors/test_converted_notebook_preprocessor.py::test_filter_include_cell
Your build timed out. Use the [--timeout=DURATION] flag to change the timeout threshold.
ERROR: (gcloud.builds.submit) build 9f2ffbf5-78f1-4602-a581-8d7fd64b8a27 completed with status "TIMEOUT"
Step #2: tests/integration/gcp/test_gcs_access.py::test_job_submission_without_gcs_access
```

Other than above error, I only see tests that are marked `PASSED`. I wonder if it is the changes in this PR that cause the timeout or something else?

abhi-g:
Were you able to merge upstream/master into your PR. If so it should have
resolved the timeout issue. If it is still persisting, it could be likely
due to some change in the PR.

vjrantal:
@abhi-g Yes, I merged latest from master.

jinchihe:
@abhi-g @karthikv2k Unluckly, seems the timeout problem can be reproduced now even if with the previous fix change., I do not think that's caused by the PR changes, but have other case still have issue or other root cause.

vjrantal:
@jinchihe @abhi-g @karthikv2k I tried to retest, but seems like I still get:

```
Your build timed out. Use the [--timeout=DURATION] flag to change the timeout threshold.
ERROR: (gcloud.builds.submit) build 07277e6b-3001-403e-80f4-3805a8b2a9ba completed with status "TIMEOUT"
```

I would be willing to continue working with the PR to make the CI job pass, but currently logs are giving relatively little actionable insights into what fails. Or, can it be that the total length of the run simply is too long with the amount of tests?

karthikv2k:
Thanks @vjrantal for your persistence! Let me spend some time today on debugging whats going on.

jinchihe:
@vjrantal just waiting for a moment, I'm fixing the problem in #353

karthikv2k:
Some random docker api timeout error in previous attempt so retesting again
/retest

karthikv2k:
@vjrantal 
I fixed the issue with the test env. Now all tests runs normally. I disabled test_job_submission_gcpmanaged because GCPManagedBackend doesn't have build_context_source in its constructor. We need to add it but for now we can disable this test. Now you need to only correct some lint errors. See the test logs for details.

karthikv2k:
CC @abhi-g ^

PR #325, opened 2019-07-29
support mounting pvc for job deployer

### LINKED ISSUE #324: Support mounting PVC easily for on-prem cluster
/kind feature

Fairing can deploy the ML trainning as the Job in the Kubernetes cluster, but for onprem cluster, that's hard to save the model after the training pod finished. For on-prem Kubernetes cluster, that's goog way to save the intermediate data and model result to the PVC, and then the following `serving` step can directly load the model and public a service (For the user cases, the two steps can use some pvc to model saving and loading).

jinchihe:
Support mount pvc to job deployer, so the trained model can be saved in the PVC derectly.

Fixes #324 

From muy initial thoughts, that can be implemented by `pod_spec_mutators`, but checked the [adding mechanism](https://github.com/kubeflow/fairing/blob/4f9e007365101443e1230ee206980ed6014f7d31/fairing/deployers/job/job.py#L53) for `pod_spec_mutators`, that's hard to pass the args, but the `pvc_name` should be pass to job, so I updated the `Job` class to let user input `pvc_name` and `pvc_mount` (optional, defaults to `/mnt`).

**Release note**:

```release-note
support mounting pvc for job deployer
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/325)

### DIFF
--- a/fairing/kubernetes/utils.py
+++ b/fairing/kubernetes/utils.py
@@ -1,4 +1,6 @@
 from kubernetes.client.models.v1_resource_requirements import V1ResourceRequirements
+from fairing.constants import constants
+from kubernetes import client
 
 def get_resource_mutator(cpu=None, memory=None):
     def _resource_mutator(kube_manager, pod_spec, namespace):
@@ -22,3 +24,24 @@ def get_resource_mutator(cpu=None, memory=None):
             else:
                 pod_spec.containers[0].resources = V1ResourceRequirements(limits=limits)
     return _resource_mutator
+
+
+def mounting_pvc(pvc_name, pvc_mount_path=constants.PVC_DEFAULT_MOUNT_PATH):
+    '''The function for pod_spec_mutators to mount persistent volume claim'''
+    mounting_name = str(constants.PVC_DEFAULT_VOLUME_NAME) + pvc_name
+    def _mounting_pvc(kube_manager, pod_spec, namespace):
+        volume_mount = client.V1VolumeMount(
+            name=mounting_name, mount_path=pvc_mount_path)
+        if pod_spec.containers[0].volume_mounts:
+            pod_spec.containers[0].volume_mounts.append(volume_mount)
+        else:
+            pod_spec.containers[0].volume_mounts = [volume_mount]
+
+        volume = client.V1Volume(
+            name=mounting_name,
+            persistent_volume_claim=client.V1PersistentVolumeClaimVolumeSource(claim_name=pvc_name))
+        if pod_spec.volumes:
+            pod_spec.volumes.append(volume)
+        else:
+            pod_spec.volumes = [volume]
+    return _mounting_pvc
--- a/tests/integration/onprem/test_pvc_mounting.py
+++ b/tests/integration/onprem/test_pvc_mounting.py
@@ -1,0 +1,115 @@
+import pytest
+import sys
+import io
+import tempfile
+import random
+import time
+import uuid
+import joblib
+import numpy as np
+
+from google.cloud import storage
+from kubernetes import client
+
+import fairing
+from fairing.constants import constants
+from fairing.kubernetes.utils import mounting_pvc
+
+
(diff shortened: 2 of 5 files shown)


karthikv2k:
Please checkout my comments on #328 .

fairing/deployers/job/job.py
```diff
@@ -24,7 +26,7 @@ class Job(DeployerInterface):
     def __init__(self, namespace=None, runs=1, output=None,
                  cleanup=True, labels=None, job_name=constants.JOB_DEFAULT_NAME,
                  stream_log=True, deployer_type=constants.JOB_DEPLOPYER_TYPE,
-                 pod_spec_mutators=None):
+                 pvc_name=None, pvc_mount_path=None, pod_spec_mutators=None):
```
karthikv2k: This should be part of the backends so that we can mount pvc in all build/train/serve jobs.

tests/integration/onprem/test_pvc_mounting.py
```diff
@@ -0,0 +1,63 @@
...
+
+    expected_result = str(uuid.uuid4())
+    fairing.config.set_deployer(deployer, namespace=namespace, cleanup=cleanup,
+                                labels={'pytest-id': expected_result}, stream_log=False,
+                                pvc_name=pvc_name, pvc_mount_path=pvc_mount_path)
+
+    remote_train = fairing.config.fn(lambda : train_fn(expected_result))
+    remote_train()
+    job_detail = str(get_job_with_labels(namespace,'pytest-id=' + expected_result))
+    assert expected_result in job_detail
```
karthikv2k: This is not actually testing for correct mounting of PVC. The train_fn should look for a dir or file in pvc. A similar example test would be the one we have for GCS access. https://github.com/kubeflow/fairing/blob/master/tests/integration/gcp/test_gcs_access.py

fairing/deployers/job/job.py
jinchihe: Updated to following the `kubenertes.utils.get_resource_mutator` implementation way that's great.

tests/integration/onprem/test_pvc_mounting.py
jinchihe: @karthikv2k Updated the tests, I do not think we can test the dir/file in the pvc since no pvc in testing env (namespace). I checked the job and deployment after deployed, and get the volume from created job (training) and deployment (serving), and assert that to ensure the pvc is mounted, as following:
```
    assert pvc_name == created_job.items[0].spec.template.spec.volumes[0].persistent_volume_claim.claim_name
    if pvc_mount_path:
        assert pvc_mount_path == created_job.items[0].spec.template.spec.containers[0].volume_mounts[0].mount_path
    else:
        assert constants.PVC_DEFAULT_MOUNT_PATH == created_job.items[0].spec.template.spec.containers[0].volume_mounts[0].mount_path
```

jinchihe:
@karthikv2k @abhi-g Any more comments? thanks.

fairing/kubernetes/utils.py
```diff
@@ -22,3 +24,24 @@ def _resource_mutator(kube_manager, pod_spec, namespace):
             else:
                 pod_spec.containers[0].resources = V1ResourceRequirements(limits=limits)
     return _resource_mutator
+
+
+def mounting_pvc(pvc_name, pvc_mount_path=constants.PVC_DEFAULT_MOUNT_PATH):
```
abhi-g: rename function to mount_pvc?

fairing/kubernetes/utils.py
```diff
@@ -22,3 +24,24 @@ def _resource_mutator(kube_manager, pod_spec, namespace):
...
+    '''The function for pod_spec_mutators to mount persistent volume claim'''
+    mounting_name = str(constants.PVC_DEFAULT_VOLUME_NAME) + pvc_name
+    def _mounting_pvc(kube_manager, pod_spec, namespace):
```
abhi-g: rename _mount_pvc.

PR #327, opened 2019-07-30
support pod_spec_mutators in Serving deployer

### LINKED ISSUE #256: How to add volumes to the serving deployer
How can we add volumes to the serving deployer?

The constructor for serving doesn't allow pod_spec_mutators to be set
https://github.com/kubeflow/fairing/blob/master/fairing/deployers/serving/serving.py

Unlike for jobs
But it looks like we can directly set the attribute.

jinchihe:
support pod_spec_mutators in Serving deployer so that user can add secret volumes to the serving deployer.

Fixes #256 

**Release note**:

```release-note
support pod_spec_mutators in Serving deployer
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/327)


abhi-g:
@jinchihe the change is looking good. Is there a unit test that is exercising this code path. If not, can you please add one.

PR #333, opened 2019-07-31
Patch

### LINKED ISSUE #332: docker created from fairing uses python2 even base docker image has python 3 as default
/kind bug

I am using Trainjob API to deploy and run it remotely on GKE cluster. 
Following is the flow:
1. After running Trainjob API it is first creating docker (working correctly)
2. Then copying all the necessary dependencies and packaging it with docker (I am still not sure about this step)
3.Then pushing docker to specified registry (working correctly)
4. Then using the same docker to run it on GKE cluster (from here the problem starts)
 In step no 4 folling is the problem
````
  File "/app/function_shim.py", line 4, in <module>
    from enum import Enum
ImportError: No module named enum
```   
I looked up the problem & following is conclusion:
1. enum package generally gives problem in python2
2. So I made sure my base docker image will use python3 as default. So should the docker generated step 2, but I think this is not the case. 

Function inside the docker should work just like the function working locally. 

Docker generated in step 2 i.e. by Faring should use python3 by default 
I think so make change in Docker command python to python3.  (below log may help)
`Docker command: ['python', '/app/function_shim.py', '--serialized_fn_file', '/app/pickled_fn.p']`



**Environment:**

- Fairing version: 0.5.3
- Kubeflow version: 0.5

Akashdesarda:
make docker run python3 as default. I faced lot of issue due to python2 which got resolved when the docker ran on python3
Fixes #332

**Release note**:

```release-note
None
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/333)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,3 +16,5 @@ All submissions, including submissions by project members, require review. We
 use GitHub pull requests for this purpose. Consult
 [GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
 information on using pull requests.
+
+-[Akash Desarda](https://github.com/Akashdesarda)
--- a/fairing/preprocessors/function.py
+++ b/fairing/preprocessors/function.py
@@ -86,7 +86,7 @@ class FunctionPreProcessor(BasePreProcessor):
 
         local_python_version = ".".join([str(x) for x in sys.version_info[0:3]])
 
-        self.command = ["python", os.path.join(self.path_prefix, FUNCTION_SHIM),
+        self.command = ["python3", os.path.join(self.path_prefix, FUNCTION_SHIM),
                         "--serialized_fn_file", payload_file_in_context, 
                         "--python_version", local_python_version]
 


CONTRIBUTING.md
```diff
@@ -15,4 +15,6 @@ again.
 All submissions, including submissions by project members, require review. We
 use GitHub pull requests for this purpose. Consult
 [GitHub Help](https://help.github.com/articles/about-pull-requests/) for more
-information on using pull requests.
\ No newline at end of file
+information on using pull requests.
+
+-[Akash Desarda](https://github.com/Akashdesarda)
```
jinchihe: Seems no need to add this?

jinchihe:
@Akashdesarda I think you can update your env to install or alias python to python3? or create a python3 virtual environment to do? Personally I think that's OK because since the `python` means `python3` in some env. Thanks.

Akashdesarda:
@jinchihe I already all things that you mentioned like using alias, python 3 venv, removing python 2 completely, still not successful

jinchihe:
> @jinchihe I already all things that you mentioned like using alias, python 3 venv, removing python 2 completely, still not successful

What'a the problem then? thanks.

karthikv2k:
Closing this pull request for now. Please discuss the problem and agree on the solution in #332 and then create a PR for it.

Issue #202, opened 2019-04-10
Which notebook image to use xgboost-high-level-apis example?


jlewi:
Is there an existing notebook image suitable for running the example:

https://github.com/kubeflow/fairing/blob/master/examples/prediction/xgboost-high-level-apis.ipynb

I tried using
gcr.io/kubeflow-images-public/tensorflow-1.13.1-notebook-cpu:v0.5.0

But it looks like its missing some dependencies.

```
import argparse
import logging
import joblib
import sys
import pandas as pd
from sklearn.metrics import mean_absolute_error
from sklearn.model_selection import train_test_split
from sklearn.impute import SimpleImputer
from xgboost import XGBRegressor
```

I was able to work around this just by installing the modules.

```
!pip install joblib
!pip install pandas
!pip install scikit-learn
```

/cc @karthikv2k @r2d4

karthikv2k:
you have to install the requirements in https://github.com/kubeflow/fairing/blob/master/examples/prediction/requirements.txt . We didn't create a separate image for this example. We will add "!pip install -r requirements.txt" as the first cell in the notebook to let the users know they have to install the deps.

PR #335, opened 2019-08-03
Fix bug in predict in examples.

### LINKED ISSUE #258: Bug in predict in the xgboost example?
https://github.com/kubeflow/fairing/blob/08c81437fd9ed720cf4b780e25097fef9fa16cd6/examples/prediction/xgboost-high-level-apis.ipynb#L175

This line seems wrong

```
return [[prediction.item(0), prediction.item(0)]]
```

Why would we want to return the same value twice? I suspect it should be

```
return [[prediction.item(0), prediction.item(1)]]
```

abcdefgs0324:
Fix bugs in the function `predict` in examples

Fixes #258 

/assign @jlewi 
/assign @zhenghuiwang

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/335)


abcdefgs0324:
Also fix #202

abcdefgs0324:
Fix #202

jlewi:
/lgtm
/approve

Thanks for the bug fix!

abhi-g:
@abcdefgs0324 please resolve integration test failure

examples/prediction/HousingServe.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, model_file=MODEL_FILE):
   def predict(self, X, feature_names):
     """Predict using the model for given ndarray."""
     prediction = self.model.predict(data=X)
-    return [[prediction.item(0), prediction.item(0)]]
+    return [[prediction.item(0), prediction.item(1)]]
```
abhi-g: @karthikv2k Is this change correct? Its possible that the prediction.item(0) was intentional as the prediction is returning a singleton, but the API requires a calls and confidence?

karthikv2k: should be returning all the results for the input X - something like return [prediction]

abcdefgs0324: Got it, thanks!

jinchihe:
@abcdefgs0324 Pls rebase, the integration test problem should have been fixed in #337 thanks.

abcdefgs0324:
@jinchihe Rebased, but it still fails.

jinchihe:
@abcdefgs0324 That's strange,  same reason. that's caused the old version of `tornado`, but in the #337 the `tornado` should has been upgraded. Need more check...

```
ImportError: cannot import name 'TracebackFuture' from 'tornado.concurrent'
```

jinchihe:
@abcdefgs0324 Can you try that by adding one line to upgrade tornado in the [Dockerfile.test](https://github.com/kubeflow/fairing/blob/master/Dockerfile.test) file?

```
pip install -U tornado
```

Dockerfile.test
```diff
@@ -24,6 +24,8 @@ RUN pip install -r examples/prediction/requirements.txt
 COPY setup.py requirements.txt /opt/fairing/
 RUN pip install -e .
 
+pip install -U tornado
```
jinchihe: Should be "RUN pip install -U tornado", thanks

abcdefgs0324:
@jinchihe Solved finally. Thanks for your help!

jinchihe:
@abcdefgs0324 Thank you for your persistence!

/lgtm

PR #351, opened 2019-08-20
option to select tfjob version + adding annotation parameter

hamedhsn:
We are running on v0.4.1 kubeflow and which is on the old version of tfjob operator. since you hardcoded the version we can not use deployer to create tfjob. This PR will provide a way to pass the version through env variable. It also adds  annotations parameter which we need to assign to the pods.

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/351)

### DIFF
--- a/kubeflow/fairing/constants/constants.py
+++ b/kubeflow/fairing/constants/constants.py
@@ -1,3 +1,5 @@
+import os
+
 TEMP_TAR_GZ_FILENAME = '/tmp/fairing.layer.tar.gz'
 DEFAULT_IMAGE_NAME = 'fairing-job'
 DEFAULT_BASE_IMAGE = 'gcr.io/kubeflow-images-public/fairing:dev'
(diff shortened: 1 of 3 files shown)


jinchihe:
@hamedhsn is that better way to let user input the tfjobs version as args, and define the default tfjobs version in the constants? if user not input, just use the default version?

hamedhsn:
we could add another parameter for the version but I thought it will be too many parameters to handle at the end. I thought it is rare that people use the old version and if they do exception will be raised with an informative message. happy to change if you think there is a better way.

jinchihe:
pls rebase and then we can test, some presubmit problem has been fixed.

jinchihe:
@hamedhsn could you please rebase to resolve conflicting so that we can test it? thanks.

hamedhsn:
@jinchihe rebased, should be good now

hamedhsn:
@jinchihe ready to merge

PR #372, opened 2019-09-13
Add service_type in serving deployer for aws

### LINKED ISSUE #371: Can not deploy serving endpoint on AWS
/kind bug

Try to use `KubeflowAWSBackend` to deploy an endpoint and meet issues. 

```
from kubeflow.fairing import PredictionEndpoint
endpoint = PredictionEndpoint(HousingServe, input_files=['trained_ames_model.dat', "requirements.txt"],
                              docker_registry=DOCKER_REGISTRY,
                              service_type='ClusterIP',
                              backend=BackendClass(build_context_source=BuildContext))
endpoint.create()
```

```
/opt/conda/lib/python3.6/site-packages/kubeflow/fairing/ml_tasks/tasks.py in create(self)
     96             self.model_class.__name__,
     97             service_type=self.service_type,
---> 98             pod_spec_mutators=self._pod_spec_mutators)
     99         self.url = self._deployer.deploy(self.pod_spec)
    100         logger.warning("Prediction endpoint: {}".format(self.url))

TypeError: get_serving_deployer() got an unexpected keyword argument 'service_type'
```

serving endpoints should be deployed successfully. 

**Environment:**

- Fairing version: (use `python -c "import kubeflow.fairing; print(kubeflow.fairing.__version__)"`): 0.5.3
- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 0.6.2
- Kubernetes version: (use `kubectl version`): 1.13.8
- OS (e.g. from `/etc/os-release`): Linux Debian

**NOTE**: If you are using fair from master, please provide us the git commit hash.

Jeffwan:

Fixes #371 

```release-note
Add service_type in serving deployer for aws
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/372)

### DIFF
--- a/kubeflow/fairing/backends/backends.py
+++ b/kubeflow/fairing/backends/backends.py
@@ -180,8 +180,10 @@ class AWSBackend(KubernetesBackend):
         pod_spec_mutators.append(aws.add_aws_credentials_if_exists)
         return Job(namespace=self._namespace, pod_spec_mutators=pod_spec_mutators)
 
-    def get_serving_deployer(self, model_class, pod_spec_mutators=None): # pylint:disable=arguments-differ
-        return Serving(model_class, namespace=self._namespace, pod_spec_mutators=pod_spec_mutators)
+    def get_serving_deployer(self, model_class, service_type='ClusterIP', # pylint:disable=arguments-differ
+                             pod_spec_mutators=None):
+        return Serving(model_class, namespace=self._namespace, service_type=service_type,
+                       pod_spec_mutators=pod_spec_mutators)
 
 class AzureBackend(KubernetesBackend):
 


kubeflow/fairing/backends/backends.py
```diff
@@ -180,8 +180,10 @@ def get_training_deployer(self, pod_spec_mutators=None):
         pod_spec_mutators.append(aws.add_aws_credentials_if_exists)
         return Job(namespace=self._namespace, pod_spec_mutators=pod_spec_mutators)
 
-    def get_serving_deployer(self, model_class, pod_spec_mutators=None): # pylint:disable=arguments-differ
-        return Serving(model_class, namespace=self._namespace, pod_spec_mutators=pod_spec_mutators)
+    def get_serving_deployer(self, model_class, service_type='ClusterIP', # pylint:disable=arguments-differ
```
jinchihe: Are you sure to set the default value to `ClusterIP`, if so the service will not public and only cluster inter can visit that, personally I suggest set to `LoadBalancer`, user can set thet to `ClusterIP` if user do not want to public that, generally the service need be visit from out of cluster.

jinchihe:
@Jeffwan Thanks for the PR, do you minder also update other backends to let user can set the service_type? thanks.

kubeflow/fairing/backends/backends.py
Jeffwan: As I think, most of the fairing users develop their model inside jupyter notebook using kubeflow. If that's the majority case, we can set default value to `ClusterIP`, users need to use `Loadbalancer` when they move their development work on laptop. 

Load Balancer will provision resources and that's also limitation at account level. That's the motivation we like to move to `ClusterIP` as default option for aws backend

Jeffwan:
> @Jeffwan Thanks for the PR, do you minder also update other backends to let user can set the service_type? thanks.

This PR mainly fixes the issue for deployer in AWS. Let's talk about default service type in https://github.com/kubeflow/fairing/pull/376

PR #376, opened 2019-09-13
Use service type ClusterIP by default

### LINKED ISSUE #373: Prediction endpoint url is wrong when using ClusterIP
/kind bug

```
from kubeflow.fairing import PredictionEndpoint
endpoint = PredictionEndpoint(HousingServe, input_files=['trained_ames_model.dat', "requirements.txt"],
                              docker_registry=DOCKER_REGISTRY,
                              service_type='ClusterIP',
                              backend=BackendClass(build_context_source=BuildContext))
endpoint.create()

```

We get endpoint results.
```
Deploying the endpoint.
Cluster endpoint: http://fairing-service-srwh2.kubeflow-anonymous.svc.cluster.local
Prediction endpoint: http://fairing-service-srwh2.kubeflow-anonymous.svc.cluster.local
```

if we run following command, it will hang there.
```
# Get sample data and query endpoint
(train_X, train_y), (test_X, test_y) = read_input("ames_dataset/train.csv")
endpoint.predict_nparray(test_X)
```

 The problem is right url should be 
`http://fairing-service-srwh2.kubeflow-anonymous.svc.cluster.local:5000/predict`

If we change to this config, prediction works well. 
```
endpoint.url='http://fairing-service-n8qv2.kubeflow-anonymous.svc.cluster.local:5000/predict'
endpoint.predict_nparray(test_X)
```

Jeffwan:

Fixes #373 #206 

Not sure if this will break post-submit test. Please correct me if I miss something I don't know. 
With this fix, clusterIP svc url can be used for prediction. If `//github.com/kubeflow/fairing/blob/master/examples/prediction/xgboost-high-level-apis.ipynb` is tested in the same kubernetes cluster, it should work. 

**Release note**:

```release-note
Use service type ClusterIP by default
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/376)

### DIFF
--- a/examples/prediction/xgboost-high-level-apis.ipynb
+++ b/examples/prediction/xgboost-high-level-apis.ipynb
@@ -304,6 +304,7 @@
    "source": [
     "from kubeflow.fairing import PredictionEndpoint\n",
     "endpoint = PredictionEndpoint(HousingServe, input_files=['trained_ames_model.dat', \"requirements.txt\"],\n",
+    "                              service_type='LoadBalancer',\n",
     "                              docker_registry=DOCKER_REGISTRY,\n",
     "                              backend=BackendClass(build_context_source=BuildContext))\n",
     "endpoint.create()"
--- a/kubeflow/fairing/backends/backends.py
+++ b/kubeflow/fairing/backends/backends.py
@@ -90,7 +90,7 @@ class KubernetesBackend(BackendInterface):
     def get_training_deployer(self, pod_spec_mutators=None):
         return Job(self._namespace, pod_spec_mutators=pod_spec_mutators)
 
-    def get_serving_deployer(self, model_class, service_type='LoadBalancer', # pylint:disable=arguments-differ
+    def get_serving_deployer(self, model_class, service_type='ClusterIP', # pylint:disable=arguments-differ
                              pod_spec_mutators=None):
         return Serving(model_class, namespace=self._namespace, service_type=service_type,
                        pod_spec_mutators=pod_spec_mutators)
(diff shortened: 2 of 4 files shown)


jinchihe:
See my comments in #372 I remember that's have long discussion for service_type before. Generally user would like to use the service as a public service by default, not ClusterIP. I think we should only make ability to let user set the service_type. agree?

jinchihe:
@Jeffwan by the way, if you changed the service_type to ClusterIP, I think the some test case prediction will be failed unless you set that as in option.

Jeffwan:
@jinchihe Hi Jinchi, I don't know the context. I looked at issue https://github.com/kubeflow/fairing/issues/206 and seems this issue mentioned fairing should not expose public endpoints. Not sure if that's still true or not. 

> "Generally user would like to use the service as a public service by default, not ClusterIP."
If this is the case, we can close above issue and I can revert changes for #206 

I need endpoint change if user specify clusterIp. This is for #373

jinchihe:
@Jeffwan I think if we update the service_type, the tests that called the `examples/prediction/xgboost-high-level-apis.ipynb` will not work, unless we also update the examples to point to the service type to Loadbalancer.

Now I'm thinking if we can use istio to route the fairing service?

Jeffwan:
> @Jeffwan I think if we update the service_type, the tests that called the `examples/prediction/xgboost-high-level-apis.ipynb` will not work, unless we also update the examples to point to the service type to Loadbalancer.
> 
> Now I'm thinking if we can use istio to route the fairing service?

Yeah. If this has to be a public endpoint, as community move to istio based routes. Ingress is definitely a better option than Loadbalancer. This should be an improvement task,

In the PR, I think maybe I can just leave this change` url = "http://{0}.{1}.svc.cluster.local:5000/predict".format(`? This is a definitely a bug blocking users to user clusterIP.  I will revert all rest changes changing to `ClusterIP`. What do you think?

jinchihe:
@Jeffwan Rechecked the code, for the `PredictionEndpoint` supports set the service_type now, I think that's OK to change the service type to `ClusterIP` as #206 logged. 

Could you also please update the service type in class `PredictionEndpoint` to `ClusterIP`, 

https://github.com/kubeflow/fairing/blob/7d7e9d15d507b4b07e50249fd6b7a5878806122f/kubeflow/fairing/ml_tasks/tasks.py#L86

and update the examples/prediction/xgboost-high-level-apis.ipynb to set the service type to `LoadBalancer`? Thanks.

Jeffwan:
@jinchihe Sure. I will make the change.

Jeffwan:
@jinchihe test pass and please have another check

PR #377, opened 2019-09-14
Fix python compatibility issue

### LINKED ISSUE #340: RuntimeError: The Python version  mismatches
/kind bug

```
DOCKERHUB_USERNAME = 'seedjeffwan'
DOCKER_REGISTRY = '{}'.format(DOCKERHUB_USERNAME)
fairing.config.set_builder('append',base_image='tensorflow/tensorflow:1.13.1-py3', registry=DOCKER_REGISTRY, push=True)
fairing.config.set_deployer('job')

def train():
    hostname = tf.constant(os.environ['HOSTNAME'])
    sess = tf.Session()
    print('Hostname: ', sess.run(hostname).decode('utf-8'))

if __name__ == '__main__':
    remote_train = fairing.config.fn(train)
    remote_train()

```

and see error message 
```
Finished upload of: seedjeffwan/fairing-job:CBF947AD
Pushed image seedjeffwan/fairing-job:CBF947AD in 113.38928063900005s.
The job fairing-job-sltq2 launched.
Waiting for fairing-job-sltq2-h4jzb to start...
Waiting for fairing-job-sltq2-h4jzb to start...
Waiting for fairing-job-sltq2-h4jzb to start...
Failed to launch fairing-job-sltq2-h4jzb, reason: Error, message: None
Cleaning up job fairing-job-sltq2...
  File "/app/function_shim.py", line 59, in <module>
    compare_version(args.python_version)
  File "/app/function_shim.py", line 38, in compare_version
    + local_python_version + ' in the local environment.')
RuntimeError: The Python version 3.5.2 mismatches with Python 3.7.3 in the local environment.

```

I create a notebook locally by running `jupyter notebook` and my machine python version is 3.7.3. When it schedule job on kubernetes, it encounter this problem. 
```
!python3 -V
Python 3.7.3
```

I enter `tensorflow/tensorflow:1.13.1-py3`  and notice python3 version is `3.5.2`. That's probably the reason.  I understand it's good to check version for potential compatibility issues. 
By looking at the code, I think it doesn't make sense to compare full version. To make sure compatibility, I think major (3) version or minor (5,6,7) are necessary, patch version is not helpful here. (See my comments below)

https://github.com/kubeflow/fairing/blob/8a350ea2c453e2e06f2b7e534de88a2936f9711a/fairing/functions/function_shim.py#L34-L38

In this example, The behavior is definitely correct but we need improvement to only compare to minor version

**Environment:**
- Fairing version: (use `python -c "import fairing; print(fairing.__version__)"`): 0.5.3
- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 0.6.0
- Minikube version: 
- Kubernetes version: (use `kubectl version`): 1.13
- OS (e.g. from `/etc/os-release`):

**NOTE**: If you are using fair from master, please provide us the git commit hash.

Jeffwan:

Fixes #340 

1. We only retrieve major and minor version for comparison in the code. 
2. Actually, this is not perfect because of potential incompatibility issue. Better to make sure remote running py version > notebook python version.  

```
Potentially compatible: Python version x.y is potentially compatible to version w.z if x == w and y < z
```
In order to strictly follow what python official doc suggest us, we need to retrieve full version and then. I can make this change if you think that's reasonable.

```
from packaging import version
if version.parse(local_python_version) > version.parse(remote_python_version):
        raise RuntimeError('The Python version ' + remote_python_version + ' mismatches \
                           with Python ' + local_python_version + ' in the local environment.')
```

**Release note**:

```release-note
Fix python compatibility issue
```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/377)

### DIFF
--- a/tests/integration/common/test_py_version_checking.py
+++ b/tests/integration/common/test_py_version_checking.py
@@ -34,17 +34,16 @@ def run_submission_with_function_preprocessor(capsys, base_image, is_match, name
         assert "mismatches" in captured.out
 
 
-def test_py_version_mismatch(capsys):
+def test_py_minor_version_mismatch(capsys):
     current_py_version = ".".join([str(x) for x in sys.version_info[0:2]])
-    if current_py_version == "3.6.5":
-        py_version = '3.5.2'
+    if current_py_version == "3.6":
+        py_version = '3.5'
     else:
-        py_version = '3.6.5'
+        py_version = '3.6'
     base_image = 'registry.hub.docker.com/library/python:{}'.format(py_version)
     run_submission_with_function_preprocessor(capsys, base_image, False)
 
-
 def test_py_version_match(capsys):
-    py_version = ".".join([str(x) for x in sys.version_info[0:3]])
+    py_version = ".".join([str(x) for x in sys.version_info[0:2]])
     base_image = 'registry.hub.docker.com/library/python:{}'.format(py_version)
     run_submission_with_function_preprocessor(capsys, base_image, True)
--- a/kubeflow/fairing/functions/function_shim.py
+++ b/kubeflow/fairing/functions/function_shim.py
@@ -44,7 +44,7 @@ def compare_version(local_python_version):
     :returns: None.
 
     """
-    remote_python_version = ".".join([str(x) for x in sys.version_info[0:3]])
+    remote_python_version = ".".join([str(x) for x in sys.version_info[0:2]])
     if local_python_version != remote_python_version:
         raise RuntimeError('The Python version ' + remote_python_version + ' mismatches \
                            with Python ' + local_python_version + ' in the local environment.')
--- a/kubeflow/fairing/preprocessors/function.py
+++ b/kubeflow/fairing/preprocessors/function.py
@@ -82,7 +82,7 @@ class FunctionPreProcessor(BasePreProcessor):
         self.output_map[temp_payload_wrapper_file] = payload_wrapper_file_in_context
 
         local_python_version = ".".join(
-            [str(x) for x in sys.version_info[0:3]])
+            [str(x) for x in sys.version_info[0:2]])
 
         self.command = ["python", os.path.join(self.path_prefix, FUNCTION_SHIM),
                         "--serialized_fn_file", payload_file_in_context,


jinchihe:
@Jeffwan Thank you! In fact I'm hesitating as appended in #340 since we cannot sure if there is potentially compatible issue.  Could you pls show the  python official doc suggestion link here for admissible evidence?

Jeffwan:
@jinchihe  I was wrong. It's not from python.org but https://docs.safe.com/fme/html/FME_Desktop_Documentation/FME_Workbench/Workbench/Python-Compatibility.htm

To be safe, python version in development environment  should be equal or less than remote python env.  This guarantees compatibility between different envs. 

Any ideas?

jinchihe:
@Jeffwan From the docs, seems only mentioned the major and minor version, does not mention the micro version, right? seems we can extend that just compile major and minor version.

/retest

Jeffwan:
> @Jeffwan From the docs, seems only mentioned the major and minor version, does not mention the micro version, right? seems we can extend that just compile major and minor version.

Right, that's what I said in current implementation we do compare micro version which is too restrict. We can just get full version and compare them. Test cases are 

Local: 3.6.8  Remote: 3.6.8 -> true 
Local: 3.6.8  Remote: 3.6.5 -> true  (even runtime has lower version, micro is same)
Local: 3.5.8  Remote: 3.6.5 -> true  (runtime supports all features in 3.5) 
Local: 3.6.8  Remote: 3.5.8 -> false  (runtime may lack of feature in 3.6)


If you agree with this, I will make another line change.

`if local_python_version != remote_python_version:` -> `if local_python_version > remote_python_version:`

jinchihe:
Let's keep this, only compare major and minor version, and igore micro version. I tested some cases, seems that's OK if major and minor version is same (for example, 3.6 or 2.7), but we cannot sure if local_python_version > remote_python_version.

/retest

Jeffwan:
@jinchihe Sounds good. Any way to check why tests failed here.?

tests/integration/common/test_py_version_checking.py
```diff
@@ -43,6 +43,14 @@ def test_py_version_mismatch(capsys):
     base_image = 'registry.hub.docker.com/library/python:{}'.format(py_version)
     run_submission_with_function_preprocessor(capsys, base_image, False)
 
+def test_py_patch_version_match(capsys):
+    current_py_version = ".".join([str(x) for x in sys.version_info[0:2]])
+    if current_py_version == "3.6.5":
```
jinchihe: That's strange, I cannot get the intergration logs, but I guess the new case failed. the case confused me, why get [0:2] but compile all version here?  and you expected the version matched? if yes, we need to pass `is_match=True` while calling `run_submission_with_function_preprocessor`

Jeffwan: Good catch. It was supposed to be [0:3] I added this one because in current implementation, we do compare micro version. Since we just compare [0:2] now, we don't need it anymore. I delete the test case and also update original test case. Please check commit and test pass now,

PR #405, opened 2019-10-25
Allow setting custom Dockerfile for ClusterBuilder

### LINKED ISSUE #404: ClusterBuilder doesn't allow setting custom Dockerfile - Needed for xgboost_synthetic
/kind bug

ClusterBuilder doesn't allow setting a docker file. Instead it always writes a given template.

https://github.com/kubeflow/fairing/blob/58c67efacc5bacbbdad3f91607cccfce4a563fd9/kubeflow/fairing/builders/cluster/cluster.py#L47

https://github.com/kubeflow/fairing/blob/7c93e888c3fc98bdf5fb0140e90f6407ce7a807b/kubeflow/fairing/builders/dockerfile.py#L9

We need this for xgboost_synthetic with 0.7.

With 0.7 our jupyter images are now built from the TF images. By default the user is set to jovyan and the user doesn't have permission to install the python packages globally.

So we need to use a custom Dockerfile that sets the user to root does the pip install and then sets it back to jovyan.

Related to kubeflow/examples#655

jinchihe:
Allow setting custom Dockerfile for ClusterBuilder by setting `dockerfile_path`, if the `dockerfile_path` is not set, will generate the Dockerfile

Fixes #404 

**Release note**:

```release-note

```

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/405)

### DIFF
--- a/kubeflow/fairing/builders/cluster/cluster.py
+++ b/kubeflow/fairing/builders/cluster/cluster.py
@@ -32,7 +32,8 @@ class ClusterBuilder(BaseBuilder):
             image_name=image_name,
             push=push,
             preprocessor=preprocessor,
-            base_image=base_image)
+            base_image=base_image,
+            dockerfile_path=dockerfile_path)
         self.manager = KubeManager()
         if context_source is None:
             raise RuntimeError("context_source is not specified")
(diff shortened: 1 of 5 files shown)


abhi-g:
@jinchi please see if you can add a test for this codepath.

@jlewi, in absence of this parameter, users would have probably created a different base image and used the base_image param. isn't that a viable option?

jlewi:
@abhi-g I think one of the main use cases for ClusterBuilder/Kaniko is building a base image. 

In the xgboost_synthetic image we are building a base image by combining a TensorFlow image with additional requirements.

abhi-g:
@jinchihe sorry I have been occupied at the Kubeflow summit. Will be able to take a look tomorrow

jinchihe:
> @jinchihe sorry I have been occupied at the Kubeflow summit. Will be able to take a look tomorrow

@abhi-g OK. Thanks. The ci testing is not stable.. The error is not caused by code change.
```
Exception: Image build failed: error creating overlay mount to /var/lib/docker/overlay2/b107e8d785e870a18f27225bf739
cef983597ed325c8a7b24f42ecc3ee71ffc4/merged: device or resource busy
```

abhi-g:
Yes, quite likely this is testing infra issue. Though since everyone is
going to be at the KF summit, I might only be able to get help tomorrow
with the infra issue.

jlewi:
/lgtm
/approve
/hold 
In case @abhi-g should take another look

Issue #315, opened 2019-07-12
Code review: possible incorrect condition


shikha130vv:
In file fairing.backends.py
In function GKEBackend
if KubeManager().secret_exists(constants.GCP_CREDS_SECRET_NAME, self._namespace):
                msg.append("It seems you don't have permission to list/access secrets in your Kubeflow cluster."
                    "We need this permission in order to build a docker image using Kubeflow cluster."
                    "Adding Kubeneters Admin role to the service account you are using might solve this issue.")

master branch after 123 commits

jlewi:
@shikha130vv Can you please explain what the incorrect condition is?

shikhabitgrit:
What I cannot understand is that if the function secret_exists returns true, which will happen if secret exists and is accessed, the message conveys that user may not have access to secrets....

xauthulei:
@shikha130vv , I have removed that conveys messages. Thanks.

jinchihe:
@shikha130vv The issue should be addressed already, would you mind close this? thanks.

PR #420, opened 2019-11-21
fix logical bug

xauthulei:
fixed issue #315 
@jinchihe ,Could you review this, thanks

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/420)


jinchihe:
@xauthulei Thanks for the contribution. But I think if Workload Identity is enabled, may not user-gcp-sa credentials secret now? I'm not sure if still need to logical now, just remove the `if`? 

comments? @abhi-g

abhi-g:
@jinchihe I think you are correct. We can remove that if condition completely. If there is no way to guess the correct builder, it's most likely not because of credentials.

xauthulei:
@jinchihe and @abhi-g ,Thanks for your great review efforts .I have remove the `if` condition and related message.

PR #471, opened 2020-03-04
Fix duplicated code in cluster builder

jinchihe:
Fix duplicated code in cluster builder. In s3_context.py and minio_context.py, defined args but does not use that, and the `push` option does not work. The PR is going to fix that, remove duplicated code.

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/471)


jinchihe:
/assign @abhi-g 

Could you please take a look? Thanks.

Issue #454, opened 2020-02-12
Better Tekton support r(possibly replace ClusterBuilder)


jlewi:
/kind feature

A common task for fairing Kubeflow is building docker images in cluster.

Right now we use ClusterBuilder and some of the other libraries to do this.

I wonder if it might not be better to offer Tekton as an option. 

In kubeflow/examples#723 I'm creating an E2E example for mnist that runs inside Jupyter on Kubeflow. This needs to build a bunch of Docker images and is currently using Cluster Builder.

Here's how this might work using Tekton

1. We provide utilities for producing a context to pass to Kaniko or some other builder
    * On GCP this amounts to creating a tarball and uploading it to GCS
1. We define in YAML the spec for the Tekton task to run Kaniko
1. We launch the Tekton task using some python utilities like apply_k8s_specs mentioned in kubeflow/fairing#453

I find the ClusterBuilder to be doing a lot of magic that can get in the way; e.g. it tries to autogenerate Dockerfiles and makes other assumptions about context.

I think it might be simpler just to give people Tekton and some other low level tools (like routines for creating contexts on GCS, S3, etc...) rather than reverse engineer the magic in fairing.

The other advantage of Tekton is that it has a lot of built in support that could be useful. For example, we can leverage Tekton's git support to easily support the use case where people build from git rather than a context.

animeshsingh:
+1. Given the overall wide array of tasks in Kubeflow which are being done using Tekton, including CI/CD, it will make sense to use it as a common tooling where we have common scenarios which require build et all be handled by Tekton, and leverage wide support of features which Tekton provides to expose Kube native resources. 

cc @jinchihe

iancoffey:
Im new to this project, but I have a bunch of experience with Tekton. Id love to be assigned this one, as a way to get better oriented with this project :)

iancoffey:
cc @jinchihe may I be assigned this one?

jinchihe:
@iancoffey sorry for later. Yeah, any process from your sides already done?  I also started investigating with @mochiliu3000 for this, as for as know, @mochiliu3000 did a design for this for reviewing now.

jinchihe:
@jlewi  In fact, I have some concerns about this:
- If so, user needs to install Tekton CRD in clusters, this may increase user time overhead and complexity, may not be good for user experience. And for some user envionment, User cannot install CRD without Cluster Admin agreement.
[...]

Issue #453, opened 2020-02-12
Create a python method to bulk apply YAML specs for KF resources


jlewi:
/kind feature

See the function `apply_k8s_specs`
https://github.com/kubeflow/examples/blob/21b24b2bf083b832d6589625d002e0865a62b183/mnist/k8s_util.py#L32

This is function is really useful in notebooks we are using as KF tutorials. In the notebook we can basically just define a list of strings containing YAML specs and then apply them. (See how its used in the mnist example in that PR).

I think it would make since to create a function like that in the fairing SDK. I didn't do that originally for reasons of expediency.

This would make it easier to use notebooks to create self paced tutorials that highlight what the K8s resources look like to help users understand how to use them.

jinchihe:
/assign 

Will investigate and implement this. Thanks.

Issue #428, opened 2019-11-28
Make new high level API using KFServing


jinchihe:
/kind feature

**Describe the solution you'd like**
Make the high level API for deploy service using kfserving. For now, there are two high Level APIs,`TrainJob` and `PredictionEndpoint` , the `PredictionEndpoint`  is using deployer `Serving`. I think we should create new high level API which calls deployer `KFServing`. That's easy for deploy inference service.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

xauthulei:
@jinchihe , I will working on this, Thanks

jinchihe:
@xauthulei Exciting! Thanks!

jlewi:
Should we get rid of the high level APIs rather than continue to expand them? 

fairing has "ml_tasks"
https://github.com/kubeflow/fairing/tree/master/kubeflow/fairing/ml_tasks

Which combine together the lower level APIS (preprocessors, builders, deployers).

In a lot of our examples, we find that this API is too constraining and its easier just to use the low level APIs directly.

Should we deprecate "ml_tasks" (e.g. PredictionEndpoint) rather than continuing to extend and expand it?

jinchihe:
@jlewi Thanks for the comments.

Let talk about `ml_tasks` in #476 

/close

Issue #234, opened 2019-04-30
Figure out python packaging for fairing


jlewi:
I think fairing is imported as 

```
import fairing
```

Should we instead make this a subpackage in "kubeflow"

e.g.

```
import kubeflow.fairing
```

jlewi:
@karthikv2k What is the plan for 0.6 ? Do you think we can/should change the import as part of 0.6?

jlewi:
@karthikv2k is thinking we will stick with `import fairing` for 0.6 because there are a lot of places that need to be updated.

jinchihe:
@jlewi Agree with you, as logged https://github.com/kubeflow/kubeflow/issues/3683, we should have a unified `kubeflow` SDK, so sub-project will import as `import kubeflow.fairing` , `import kubeflow.kfp` and `import kubeflow.lfmd` etc ...

jlewi:
So should we change to kubeflow.fairing in 0.6?

swiftdiaries:
Yeah, +1

jinchihe:
@jlewi Guess you want to @ me? :-) I will take this.
/assign @jinchihe

jinchihe:
hello @jlewi @animeshsingh 

After thinking, there are two proposals:

## Option 1:  Radical way
1. Rename the project name to `kubeflow-sdk` or `kubeflowsdk` (or better name?).
2. The pacakage name will be `kubeflow` (in Pypi), user will install by `pip install kubeflow`
3. Make new folder named `kubeflow`. And move all current `fairing` files to the create `kubeflow` as subpackage, we also can sync regularly from other components later to kubeflow (such as `kubeflow.kfp` and `kubeflow.lfmd`, `kubeflow.kfserving`)
4. Update the `setup.py` files and other python file (imports to relative import way). 
5. Update the samples to import fairing as `from kubeflow import fairing`, includes other repo such as `kubeflow examples`.

**Pros** Radical way, we can build unified kubeflow sdk package.
**Cons**: Changes are big.

## Option 2: Conservative way
1. Just update required file from fairing to `kubelow.fairing` such as in setup.py and __init__ files, and import statements.
2. Update all samples by `import kubeflow.fairing...` and other repos.

**Pros**: changes is small.
**Cons**: seems so hack this way.

Personally I prefer to option 1, since the option 2 is so hack and Option 1 can let us build unified sdk package, user no need to install python package for each components that may let user confusing or unhappy.

[...]

PR #476, opened 2020-03-10
Update roadmap for 2020

### LINKED ISSUE #468: fairing 2020 ROADMAP
/kind feature

The roadmap for fairing is probably in need of a refresh. What are the high level goals for 2020?

@jinchihe @animeshsingh any interest in driving this? I think it would be great to put together some ideas and discuss it at a community meeting.

jinchihe:
Update roadmap for 2020

Fixes #468

---
This change is [Reviewable](https://reviewable.io/reviews/kubeflow/fairing/476)

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -1,11 +1,17 @@
-# Q2 2019
-1. Distirbuted trianing TF (1.x and 2.0)
-1. Distirbuted training for LightGBM
-1. TF-Serving as an option for delpolyment in addition to Seldon model server
-1. Improve documentation and testing
-1. Improved support for Append builder
+# 2020 Roadmap
 
-# Q3 and Q4 2019 Big Rocks
-1. Support for distributed training in other frameworks: PyTorch, Spark ML, etc.
-1. Integrations: KF-Serving, KF-Pipelines, Katib
-1. Integration tests for on-prem use case
+1. Fairing 1.0 release (P1, 2020 Q2)
+   - Improve Fairing quality for 1.0 production readiness, and enhance testing and examples.
+   - Enhancement [API Documents](https://kubeflow-fairing.readthedocs.io/en/latest/index.html).
+   - Deprecating high level API `ml_tasks`, maybe warn the deprecating in 1.0 release.
+
+2. Fairing provides a python SDK for Kubeflows components
+   - Support Tekton (P1, 2020 Q2).
+   - More generic interacts with Kubernetes such as bulk apply YAML Specs (P1, 20202 Q2).
+   - Support more builder such as podman or s2i and more backends such as IBM Cloud (P2, 20202 Q2-Q3)
+   - Intergration with other kubeflow components such as Katib and more frameworks etc. (P2, 2020 Q3-Q4).
+   - Add istio routing rules in Fairing (P2, 2020 Q3-Q4)
+
+3. Generate one single SDK for Kubeflow components (P2 2020 Q3-Q4).
+   - Consider and make an auto mechanism to generate Kubeflow master SDK (maybe better to name `kubeflow`).
+   - Reconcile the APIs of kubeflow components and satisfy ML logical manner.


jinchihe:
/cc @jlewi @animeshsingh @abhi-g 

I planed the roadmap for 2020, some items are from the issue under Fairing. Would you please take a look? Thanks a lot!

roadmap.md
```diff
@@ -1,11 +1,14 @@
...
+# 2020 Roadmap
+1. Supports local podman builder to submit jobs to CRI-O runtime [#472](https://github.com/kubeflow/fairing/issues/472)
+1. Support more backends cluster, such as IBM Cloud
+1. Add a GCP function to generate stackdriver links [#455](https://github.com/kubeflow/fairing/issues/455)
+1. Better Tekton support (possibly replace ClusterBuilder) [#454](https://github.com/kubeflow/fairing/issues/454)
+1. Make new high level API using KFServing [#428](https://github.com/kubeflow/fairing/issues/428)
+1. Create a release process doc and a way to run tests on wheel package [#250](https://github.com/kubeflow/fairing/issues/250)
+1. Add istio routing rules in Fairing [#208](https://github.com/kubeflow/fairing/issues/208)
+1. Clean up examples
+1. Integration with Pipeline and katic etc.
```
andreyvelich: Should it be Katib?

jinchihe: Done. Thanks for reviewing @andreyvelich

jlewi:
Thanks this is great.

Any thoughts about fairing 1.0?

Any plans to rethink what the objectives and overall structure of fairing should be? Should we have a high-level objective to rethink the overall API?

jlewi:
Could you please present the roadmap at an upcoming community meeting?

jinchihe:
Thanks @jlewi I'm on the way now, not sure if can access network later, if cannot I will show that in next week meeting, thanks.

jinchihe:
> Any plans to rethink what the objectives and overall structure of fairing should be? Should we have a high-level objective to rethink the overall API?

@jlewi Would you please elaborate about this ? Thanks!

jlewi:
What is the objective and scope of fairing? I think if we define this that will help prevent scope creep and fairing from becoming a random pile of python code.

I think when fairing started one of its goals was to create a more an intuitive experience for data scienctists who were familiar with python but not K8s.

This was the motivation for creating [ml_tasks](https://github.com/kubeflow/fairing/tree/master/kubeflow/fairing/ml_tasks). 

Another objective was to define an API that could have multiple backends (e.g. either Kubeflow or Cloud services). That's why there's code for deploying models on [GCP managed services](https://github.com/kubeflow/fairing/blob/master/kubeflow/fairing/deployers/gcp/gcpserving.py). Its a bit unclear what the future for such code is.

Specifying the objective and scope would help us make decisions about what to do about this and other features.

My initial thought would be to try to narrow down the scope of fairing. Here's an initial proposal

**Objective**: Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.

One of the primary use cases I see for SDK is letting people interact with Kubeflow & kubernetes without having to shell out to kubectl.

A primary use case is our tutorial notebooks (e.g. [mnist](https://github.com/kubeflow/examples/tree/master/mnist)). When doing a walk through in a notebook, it would be nice to use idiomatic python rather than shell out to kubectl.

An example of how fairing would help would be creating a bulkApply method (#453) that allows people to easily apply K8s objects from notebooks or other python code.

That's also consistent with the idea of trying to use Tekton (#454) for doing the heavy lifting of building containers and just building a light-weight SDK that makes it easier to interact with Tekton using idiomatic python.

If the objective is to create a light-weight, thin python SDK then that suggests we should start deprecating some of the higher level APIs (e.g. ml_tasks) and not creating new high level APIs (e..g for KFServing #428).

Does fairing need to define higher level APIs? That seems redundant to me. In my view KFServing is the high level API. So the SDK should just be some syntactic sugar ontop of that mainly aimed at creating an idiomatic python experience.

I think if we narrow the scope of fairing that will make it easier to maintain and graduate to 1.0.

Thoughts?

/cc @animeshsingh to get his thoughts

jinchihe:
@jlewi Great thanks for the comments! Jeremy. Overall agree with you.

- Will implement a bulkApply method (#453) to interact with kubeflow/k8s. Will add this into Roadmap.
- Drop the item about implementing high level API for KFServing (#428). Will remove this from Roadmap.
- Will take high priority to investigate and integration with Tekton (#454).
- In fact I'm hesitating if we should remove the `ml_tasks`, as we cannot sure if there are someone are using the API, but failure if upgrading fairing. Personally we should keep compatibility. Comments?

animeshsingh:
> I think when fairing started one of its goals was to create a more an intuitive experience for data scienctists who were familiar with python but not K8s.
> 
> This was the motivation for creating [ml_tasks](https://github.com/kubeflow/fairing/tree/master/kubeflow/fairing/ml_tasks).
> 
> Another objective was to define an API that could have multiple backends (e.g. either Kubeflow or Cloud services). That's why there's code for deploying models on [GCP managed services](https://github.com/kubeflow/fairing/blob/master/kubeflow/fairing/deployers/gcp/gcpserving.py). Its a bit unclear what the future for such code is.
> 
> Specifying the objective and scope would help us make decisions about what to do about this and other features.
> 
> My initial thought would be to try to narrow down the scope of fairing. Here's an initial proposal
> 
> **Objective**: Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.
> 
> One of the primary use cases I see for SDK is letting people interact with Kubeflow & kubernetes without having to shell out to kubectl.
> 
> A primary use case is our tutorial notebooks (e.g. [mnist](https://github.com/kubeflow/examples/tree/master/mnist)). When doing a walk through in a notebook, it would be nice to use idiomatic python rather than shell out to kubectl.
> 
> An example of how fairing would help would be creating a bulkApply method (#453) that allows people to easily apply K8s objects from notebooks or other python code.
> 
> That's also consistent with the idea of trying to use Tekton (#454) for doing the heavy lifting of building containers and just building a light-weight SDK that makes it easier to interact with Tekton using idiomatic python.
> 
> If the objective is to create a light-weight, thin python SDK then that suggests we should start deprecating some of the higher level APIs (e.g. ml_tasks) and not creating new high level APIs (e..g for KFServing #428).
> 
> Does fairing need to define higher level APIs? That seems redundant to me. In my view KFServing is the high level API. So the SDK should just be some syntactic sugar ontop of that mainly aimed at creating an idiomatic python experience.
> 
> I think if we narrow the scope of fairing that will make it easier to maintain and graduate to 1.0.
> 
> Thoughts?
> 
> /cc @animeshsingh to get his thoughts

@jlewi  TBH, I have always struggled to understand the value proposition of Fairing overall. Is there adoption? How many folks are using it? If its a Python SDK for ML tasks and multiple cloud providers backed by Tekton, why can't KFP be that SDK. It has support for all Kubeflow components, can deploy to multiple cloud providers, and can be backed by Tekton behind KFP as well....

nrchakradhar:
@animeshsingh As jlewi mentioned in the meeting, the low level APIs related to preprocessing, container building are useful. In its current form also it's useful. After adding podman and tekton it will be even more useful. The output of fairing is being used in KFP in our use cases.

The high level APIs are not being used currently.
If there is a view to combine pipelines and fairing into kubeflow SDK, it will also be good.

nrchakradhar:
Forgot to mention that backend API is also being used and is quite flexible for adapting easily to our specific installations. We have custom backend for the same which is also being used and there has been positive feedback on this.

jlewi:
@animeshsingh I think if we define the objective then that help answers whether the SDK should be one and the same with the pipelines SDK. My suggestion was

**Objective**: Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.

If this is the objective then I think it follows why the SDK should be separate from the pipelines SDK.

Consider the specific use case "How do you submit a TFJob from Python."?

Absent fairing there's a couple options

1. Shell out to `kubectl`
1. Use the Kubernetes Python Client
1. Use the Pipelines Python SDK

Shelling out to `kubectl` from python is bad (hopefully we agree on that).

The K8s python client is pretty good IMO but it has some rough edges when dealing with CRs; e.g. people have to specify the Kind, Version, etc.... We don't have type hints or other intelli sense etc... I think there's room for improvement; e.g. some light weight syntactic sugar (ideally automatically code generated so it works for all CRs).

Using the pipelines SDK to submit a TFJob is not "lightweight". If we use the pipelines SDK to generate and run a 1 step pipeline to submit a TFJob this is fundamentally different then just running `kubectl apply -f mytfjob.yaml`.

There are some folks that might choose to go down this route of making everything a pipeline. However, I think there will be enough demand for a light-weight python SDK that we should invest in that.

Also in my mind I tend to think of the Kubeflow Python SDK as a collection of python modules. e.g.

```
import kubeflow.pipelines
import kubeflow.k8s
import kubeflow.kfserving
import kubeflow.util
``` 

as opposed to a monolithic repo/package.

There will be some modules that are application specific (e.g. the pipelines compiler). There will also be utilities that are not application specific (e.g. a bulkApply method or container utilities).

Utilities/functionality which are not application specific should probably be defined outside of SDKs for specific applications and reused.

If we buy into that could add an objective along the lines of

Objective: Provide low level python utilities and tools which are used to build application specific python SDKs. 

> In fact I'm hesitating if we should remove the ml_tasks, as we cannot sure if there are someone are using the API, but failure if upgrading fairing.

I don't think we should immediately delete or remove code. I think the point of the roadmap is to provide direction about where we are heading to developers and users. So the point would be to convey to developers we don't want to continue to invest and expand ml_tasks because we don't think that is the right solution. Similarly for users we want to tell users that they should not use or stop using ml_tasks and instead point them in another direction.

jinchihe:
> Forgot to mention that backend API is also being used and is quite flexible for adapting easily to our specific installations. We have custom backend for the same which is also being used and there has been positive feedback on this.

Thanks @nrchakradhar I think we need keep that :-)

jinchihe:
@jlewi Already updated as your comments. Would you please take a look and let this check-in? Thanks a lot!

jlewi:
Thanks @jinchihe Can we make the ROADMAP a bit higher level? All of the issues you list are great and I'm glad to see they are being prioritized.  They feel a bit tactical though  (eg. https://github.com/kubeflow/fairing/issues/453). I think a ROADMAP should be a bit higher level.

Can you think of 2-3 big rocks or themes which the various items you list could roll up to?

I think "Production Readiness" or "Fairing 1.0" would be a great rollup for a lot of the fit and finish work (e.g. cleanup examples, releases, testing).

animeshsingh:
> @animeshsingh I think if we define the objective then that help answers whether the SDK should be one and the same with the pipelines SDK. My suggestion was
> 
> **Objective**: Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.

Totally agree

> If this is the objective then I think it follows why the SDK should be separate from the pipelines SDK.
> 
> Consider the specific use case "How do you submit a TFJob from Python."?
> 
> Absent fairing there's a couple options
> 
> 1. Shell out to `kubectl`
> 2. Use the Kubernetes Python Client
> 3. Use the Pipelines Python SDK
> 
> Shelling out to `kubectl` from python is bad (hopefully we agree on that).
> 
> The K8s python client is pretty good IMO but it has some rough edges when dealing with CRs; e.g. people have to specify the Kind, Version, etc.... We don't have type hints or other intelli sense etc... I think there's room for improvement; e.g. some light weight syntactic sugar (ideally automatically code generated so it works for all CRs).
> 
> Using the pipelines SDK to submit a TFJob is not "lightweight". If we use the pipelines SDK to generate and run a 1 step pipeline to submit a TFJob this is fundamentally different then just running `kubectl apply -f mytfjob.yaml`.
> 

Agreed, and this can go back to another usecase for KFP we have heard in the community, which is can we allow a "lightweight" kfp which can run on your laptop, maybe using compose, or python processes

> Also in my mind I tend to think of the Kubeflow Python SDK as a collection of python modules. e.g.
> 
> ```
> import kubeflow.pipelines
> import kubeflow.k8s
> import kubeflow.kfserving
> import kubeflow.util
> ```
> 
> as opposed to a monolithic repo/package.

Agreed, and in my opinion the base way for Fairing would to take inspiration from MLFlow SDK, and bring a single common SDK to interact with all of Kubeflow components, plus utilities as you mention like exposing common Kube tasks etc.

> There will be some modules that are application specific (e.g. the pipelines compiler). There will also be utilities that are not application specific (e.g. a bulkApply method or container utilities).
> 
> Utilities/functionality which are not application specific should probably be defined outside of SDKs for specific applications and reused.
> 
> If we buy into that could add an objective along the lines of
> 
> Objective: Provide low level python utilities and tools which are used to build application specific python SDKs.
> 

So IMHO, we shall take an inspiration from MLFlow SDK, and model Fairing along the similar lines, but also bring some abstract level concepts around Kubeflow like "projects" etc, which can be used to collect and group all the necessary artifacts like data, initial model code etc before a user enters into ML lifecycle through Kubeflow

@jlewi

jinchihe:
@jlewi Thanks a lot! Updated as your comments.
Roll up 2 big rocks, and drop some detailed items, and proposed priority for those ietms.

jlewi:
Thanks.

/lgtm
/approve

/hold for @animeshsingh 

@animeshsingh look good to you?

jinchihe:
>>  Also in my mind I tend to think of the Kubeflow Python SDK as a collection of python modules. e.g.

    >> import kubeflow.pipelines
    >> import kubeflow.k8s
    >> import kubeflow.kfserving
    >> import kubeflow.util

    as opposed to a monolithic repo/package.

> Agreed, the best goal for Fairing should be to bring a single common SDK to interact with all of Kubeflow components, plus utilities as you mention like exposing common Kube tasks etc.

@jlewi @animeshsingh Today I read your comments again after talked with Animesh. Do you think we should implememt a Kubeflow Python SDK to include all components under Kubeflow? if we want to ceate big one, we should rename the SDK. 

I once suggested that to make a single and big one SDK which includes all components, but was persuaded in the https://github.com/kubeflow/fairing/issues/234#issuecomment-528190658 :-)

Now Here's what we did for that by using [namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/) to make separation of SDK for each component such as TFJOB and KFServing etc ... 

Now user can install SDK of each components and then import by 
```
import kubeflow.tfjob
import kubeflow.kfserving
import kubeflow.pytorchjob
```
Personally most import user scenarios of Fairing is that:  User has ML training code, user just want to train on kubernetes, then Fairing is great tool to easily build a image and quick create Job/TFJob to training remotely, then deploy an inference service if needed. We should not drop the target. We can intergrate all kubeflow components in Fairing, so that user can use python to finish all steps of ML.

animeshsingh:
Thanks @jinchihe . So essentially I am hoping this becomes the roadmap becomes the northstar for Fairing, and it maybe harder to get in particular release cycles, but want to ensure the direction is captured and reflected in the doc. 

In that spirit, I think the best way is again to expand the scope of Fairing to be "the SDK for Kubeflow", with a thin layer which can work with all Kubeflow components. For e.g. let's take "Training". The way this can be structured is

import kubeflow.training.tfjob
import kubeflow.training.pytorchjob
import kubeflow.training.mpijob
import kubeflow.training.hpo (or katib)...

Similary for serving

import kubeflow.serving.kfserving
import kubeflow.serving.tfserving (if we are supporting it still)

....

jlewi:
> way is again to expand the scope of Fairing to be "the SDK for Kubeflow"

What does this really mean? Does it just mean that fairing is a catch all for "Kubeflow"?

Put another way based on this definition what would be defined as out of scope for fairing?

How are we going to avoid and resolve debates between all the different application owners and working groups?

Here's an example

* Suppose the training and KFServing folks decide they want a higher level python API (similar to SKLearn) that bundles training and inference into one higher level primitive
* Suppose the KFP folks are totally opposed to that because their view is the way to bundle things is to construct a multi-step workflow

How would we resolve that?

I think one way to resolve/prevent that is to narrowly define scope. Even what I proposed 

> Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.

is arguably too broad. The intent at least with "thin python SDK" is to say that defining higher level APIs is out of scope. 

> One way to accelerate this can be that once we have defined the plug points and abstractions, we can go to individual component owners to ensure they can plug-in their component to fairing.

Why do we need to define new plug in points and abstractions? Kubernetes already provides us a well defined resource model that we have consensus on.

I'm skeptical that we will be able to achieve consensus on higher level abstractions. Why not just lean into that and assume there will be more than one oppinionated high level SDKs?

I think the more narrowly define scope the easier it will be for people to decide whether to get on board or go their own way.

Off hand it seems like fairing is providing utility functions for three purposes

* Working with notebooks (e.g. preprocessing) them
* Building containers
* Applying and interacting with k8s resources

So we could potentially scope down fairing to those areas.

animeshsingh:
@jlewi 

> > way is again to expand the scope of Fairing to be "the SDK for Kubeflow"
> 
> What does this really mean? Does it just mean that fairing is a catch all for "Kubeflow"?
> 
> Put another way based on this definition what would be defined as out of scope for fairing?
> 

Yes. Just like kfctl is positioned as a tool for deploying and managing Kubeflow, Fairing becomes the toolkit/SDK for "using" Kubeflow. And the components it supports should be the major components of Kubeflow which are already publishing their SDKs. In that case, Fairing becomes responsible to tie them together and provide a single one.

> How are we going to avoid and resolve debates between all the different application owners and working groups?
> 
> Here's an example
> 
> * Suppose the training and KFServing folks decide they want a higher level python API (similar to SKLearn) that bundles training and inference into one higher level primitive
> * Suppose the KFP folks are totally opposed to that because their view is the way to bundle things is to construct a multi-step workflow
> 
> How would we resolve that?
> 

Earlier on there was a proposal to define a single SDK for Kubeflow - this is an extension of that. As of now, every component owner is responsible for ensuring there is a Pipeline component which needs to be created for interacting with that component. And most components are adhering to that. Similarly, if we make a "python sdk"  a recommended practice for every component, we can then plug in Fairing/

> I think one way to resolve/prevent that is to narrowly define scope. Even what I proposed
> 
> > Fairing is a thin python SDK that lets people interact with Kubeflow and Kubernetes using idiomatic python.
> 
> is arguably too broad. The intent at least with "thin python SDK" is to say that defining higher level APIs is out of scope.
> 
> > One way to accelerate this can be that once we have defined the plug points and abstractions, we can go to individual component owners to ensure they can plug-in their component to fairing.
> 
> Why do we need to define new plug in points and abstractions? Kubernetes already provides us a well defined resource model that we have consensus on.
> 

So "abstractions" is a wrong word to use here, but by plug points I mean if there are certain guidelines which different application area owner should follow for their Python SDK, which will allow easy integration into Fairing, we should publish that.

jinchihe:
@jlewi and @animeshsingh Great thanks for your comments. According to your points, I think we can have a neutral solution:
- Fairing intergrate SDK under kubeflow and keep [namespace packages](https://packaging.python.org/guides/packaging-namespace-packages/) to make separation of SDK for each component so that user can:
```
from kubeflow import kfserving
from kubeflow import tfjob
from kubeflow import matadata
....
```

- Update Faring [config](https://github.com/kubeflow/fairing/blob/master/kubeflow/fairing/config.py) as kubeflow python Client after intergration.

Currently that's like below that I think that's not best.
```
fairing.config.set_builder(...)
fairing.config.set_deployer(...)
fairing.config.run()
```
After updated,  user can use the SDK as kubeflow python Client, like this as Animesh suggested, seems more clear:
```
From kubeflow.fairing import Client as kubeflow_client
kubeflow_client.build(...)
kubeflow_client.training.tfjob(...) or  kubeflow_client.training(framework='tfjob', ...) 
kubeflow_client.deploy.kfserving(...) or kubeflow.kfserving(...)
```

Also we should support interacting with Kubeflow and Kubernetes, user can keep in Python/notebooks. no need to run command line such as kfctl or oc.

Is that good? Thanks.

jlewi:
@jinchihe I like your proposal about package namespacing. Does that address the question though of what's in scope for fairing?

@jinchihe and @animeshsingh looking at @jinchihe 's proposal

```
From kubeflow.fairing import Client as kubeflow_client
kubeflow_client.build(...)
kubeflow_client.training.tfjob(...) or  kubeflow_client.training(framework='tfjob', ...) 
kubeflow_client.deploy.kfserving(...) or kubeflow.kfserving(...)
```

Can we define the problems that we are trying to solve with `build`, `tfjob`, `kfserving`?

All three of those correspond to K8s objects. (For `build` we could use a Tekton task).

So out of the box today those objects can be created using the Kubernetes client library. Its not idiomatic python and unless we get code generation for CRs working users have to use the Kubernetes REST client library and specify low level information like kind, version, etc...

Suppose we get python client generation working for all KF custom resources; what are the remaining problems that the SDK would be addressing?

We could probably identify various helper methods that would be applicable across application; e.g.

* bulkApply
* build/buildAndUpload  Docker context (to GCS, S3, etc...) for use with Tekton for container builds
* waitForCondition
* FetchLogs

What else is needed? In particular, do you think that auto-generated python client libraries are insufficient? Assuming we do; can we create a multi-level SDK? e.g. lowest level consists of auto-generated client libraries and then higher (hand crafted) libraries are built on top of that? (I think this is what sagemaker does).

What would be the guiding principle(s) behind a high level libraries for every KF application? 

If custom code is required for every application how are we going to sustain that?

animeshsingh:
> @jinchihe and @animeshsingh looking at @jinchihe 's proposal
> 
> ```
> From kubeflow.fairing import Client as kubeflow_client
> kubeflow_client.build(...)
> kubeflow_client.training.tfjob(...) or  kubeflow_client.training(framework='tfjob', ...) 
> kubeflow_client.deploy.kfserving(...) or kubeflow.kfserving(...)
> ```
> 
> Can we define the problems that we are trying to solve with `build`, `tfjob`, `kfserving`?
> 
> All three of those correspond to K8s objects. (For `build` we could use a Tekton task).
> 
> So out of the box today those objects can be created using the Kubernetes client library. Its not idiomatic python and unless we get code generation for CRs working users have to use the Kubernetes REST client library and specify low level information like kind, version, etc...
> 
> Suppose we get python client generation working for all KF custom resources; what are the remaining problems that the SDK would be addressing?

KFServing and TF Job already have Python SDKs we have generated

> We could probably identify various helper methods that would be applicable across application; e.g.
> 
> * bulkApply
> * build/buildAndUpload  Docker context (to GCS, S3, etc...) for use with Tekton for container builds
> * waitForCondition
> * FetchLogs
> 
> What else is needed? In particular, do you think that auto-generated python client libraries are insufficient? Assuming we do; can we create a multi-level SDK? e.g. lowest level consists of auto-generated client libraries and then higher (hand crafted) libraries are built on top of that? (I think this is what sagemaker does).

Even if we "plug-in" the existing SDKs as is, without adding any top level functionalities, just that fact that installing Fairing gives me access to all the SDKs, in a more conceptualized logical manner (training, serving, pipelines, metadata...) is a great win

Would have @jinchihe comment if a high level library is needed for these

animeshsingh:
@jinchihe is out on PTO for a while - once he is back we can conclude and close this

jlewi:
> KFServing and TF Job already have Python SDKs we have generated

Thanks for the reminder this is really great.  Maybe rather than try to resolve it in this PR we should add to the ROADMAP an item to reconcile the APIs and figure out what's needed beyond the auto-generated client libraries.

> just that fact that installing Fairing gives me access to all the SDKs, in a more conceptualized logical manner (training, serving, pipelines, metadata...) is a great win

I agree that packaging them all up into a single "pip install" command is nice and something we should do.

jinchihe:
> > KFServing and TF Job already have Python SDKs we have generated
> 
> Thanks for the reminder this is really great. Maybe rather than try to resolve it in this PR we should add to the ROADMAP an item to reconcile the APIs and figure out what's needed beyond the auto-generated client libraries.
> 
> > just that fact that installing Fairing gives me access to all the SDKs, in a more conceptualized logical manner (training, serving, pipelines, metadata...) is a great win
> 
> I agree that packaging them all up into a single "pip install" command is nice and something we should do.

OK, I'm going to add one items for this in ROADMAP. But for now I think we need take high priority to implement Tekton support and bulkApply etc...

jlewi:
thanks!

/lgtm

PR #483, opened 2020-04-09
Add functions to apply Kubernetes objects

### LINKED ISSUE #453: Create a python method to bulk apply YAML specs for KF resources

jinchihe:
Add functions to apply Kubernetes objects
```
from kubeflow.fairing.kubernetes.manager import KubeManager
kubeflow_client = KubeManager()

spec = '''...'''
specs = [ ... ]

kubeflow_client.apply_namespaced_object(spec, 'create|patch|replace|delete')
kubeflow_client.apply_namespaced_objects(specs, 'create|patch|replace|delete')
```

Fixes #453

### DIFF
--- a/kubeflow/fairing/kubernetes/manager.py
+++ b/kubeflow/fairing/kubernetes/manager.py
@@ -1,5 +1,6 @@
 import logging
 import retrying
+import yaml
 
 from kubernetes import client, config, watch
 from kfserving import KFServingClient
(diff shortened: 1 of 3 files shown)


jinchihe:
/assign @abhi-g 
/cc @jlewi 

That's ready for reviewing, would you please take a look? Thanks!

abhi-g:
/lgtm
/approve

Thanks @jinchihi.

PR #490, opened 2020-04-29
Use single snapshot to reduce build times

iancoffey:

I have noticed that, when building with the cluster resource with large image layers, the multiple snapshots that are taken can add a **lot** amount of time to the build. You will  see these as the *Taking snapshot of full filesystem...* logs in the builder logs.

Using a single snapshot at the end of the process should shave quite a bit of time off the build/

**Release note**:
```release-note
Reduce cluster build time with Kaniko by making use of "--single-snapshot" ability of executor.
```


iancoffey:
@googlebot I fixed it

iancoffey:
@googlebot I fixed it

jinchihe:
/lgtm
/approve

Thanks @iancoffey

PR #493, opened 2020-05-12
bump KANIKO_IMAGE to include important fixes

iancoffey:
There are fixes for these nasty apt-based bugs included into the current release. For instance, `BentoML` generated Dockerfiles cant build on current default `0.14.0`.

https://github.com/GoogleContainerTools/kaniko/issues/940
https://github.com/GoogleContainerTools/kaniko/issues/793

Updates Kaniko Executor default image to v0.22.0.

**Release note**:

```release-note
Updates Kaniko Executor default image to v0.22.0.

```

### DIFF
--- a/kubeflow/fairing/builders/cluster/minio_context.py
+++ b/kubeflow/fairing/builders/cluster/minio_context.py
@@ -4,7 +4,7 @@ from kubeflow.fairing.kubernetes.manager import client, KubeManager
 from kubeflow.fairing import utils
 from kubeflow.fairing import constants
 
-constants.constants.KANIKO_IMAGE = "gcr.io/kaniko-project/executor:v0.14.0"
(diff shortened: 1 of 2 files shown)


kubeflow/fairing/builders/cluster/minio_context.py
```diff
@@ -4,7 +4,7 @@
 from kubeflow.fairing import utils
 from kubeflow.fairing import constants
 
-constants.constants.KANIKO_IMAGE = "gcr.io/kaniko-project/executor:v0.14.0"
+constants.constants.KANIKO_IMAGE = "gcr.io/kaniko-project/executor:v0.22.0"
```
jinchihe, line 7: @iancoffey Could you please enhance here, do not define here (remove the `constants.constants.KANIKO_IMAGE`), directly import the value that's definied in `constants`? Thanks.

iancoffey: removed that dupe, thanks

PR #495, opened 2020-05-19
Updated manager.py for reading secrets to check for the existence of …

raviranjan0309:
**Issue:**
When someone runs Kubeflow Faring from Kubeflow Pipeline, he/she will get below error:
**IOError: [Errno 2] No such file or directory: ‘/etc/secrets/user-gcp-sa.json’**

**Reason:**
When Kubeflow Fairing tries to launch the pod for model training like LightGBM distributed parallel training, it tries to check the presence of the secret by using list instead of get and when list fails due to lack of permission, it just does not attach any secrets to the pod. This causes the whole Kubeflow pipeline to fail.

**Description of your changes:**
The secret_exists function is incorrectly using list_namespaced_secret (https://github.com/kubeflow/fairing/blob/e0d44e870b467bbd773f836a8b1b648b79019dd1/kubeflow/fairing/kubernetes/manager.py#L220) to check for the existence of a secret instead of read_namespaced_secret. 
The default pipeline-runner SA used by Kubeflow Pipelines does not have the list permission for [secrets](https://github.com/kubeflow/manifests/blob/9e6ef8681003c195560fb7b3b75211830c19b7a0/pipeline/pipelines-runner/base/cluster-role.yaml#L9), so fairing fails to check that the secret exists and fails to attach it. kubeflow/pipelines#3742
This helped me run Kubeflow fairing from the Kubeflow pipeline successfully.

### DIFF
--- a/kubeflow/fairing/kubernetes/manager.py
+++ b/kubeflow/fairing/kubernetes/manager.py
@@ -217,7 +217,7 @@ class KubeManager(object):
         :returns: bool: True if the secret exists, otherwise return False.
 
         """
-        secrets = client.CoreV1Api().list_namespaced_secret(namespace)
+        secrets = client.CoreV1Api().read_namespaced_secret(namespace)
         secret_names = [secret.metadata.name for secret in secrets.items]
         return name in secret_names
 


raviranjan0309:
/assign @jinchihe

Please look into my PR and let me know for more details

raviranjan0309:
@karthikv2k @jinchihe Please review this pull request. I am working on building ML platform using Kubeflow. This change is crucial to run Kubeflow fairing from the pipeline.

raviranjan0309:
@jinchihe I see the test case failed. I just made a single line code change and checked-in the same. Do I need to perform any extra step to make this test run success?

kubeflow/fairing/kubernetes/manager.py
```diff
@@ -217,7 +217,7 @@ def secret_exists(self, name, namespace):
         :returns: bool: True if the secret exists, otherwise return False.
 
         """
-        secrets = client.CoreV1Api().list_namespaced_secret(namespace)
+        secrets = client.CoreV1Api().read_namespaced_secret(namespace)
```
jinchihe, line 220: The `read_namespaced_secret` function should be like this `read_namespaced_secret(name, namespace, pretty=pretty, exact=exact, export=export)` from docs https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#read_namespaced_secret. The `name` should be required args.

jinchihe:
@raviranjan0309 we need to ensure the CI test pass. And I'm not sure what the different permission requirement between `list_namespaced_secret ` and `read_namespaced_secret`? if you are sure that works fine, would you please show more? Thanks.

jinchihe:
@raviranjan0309 Would you please update this PR? Thanks.

raviranjan0309:
@jinchihe I am working on updating secrets permissions in pipeline runner under Kubeflow manifest. If that works then there will not be any need to make the change here.

jinchihe:
> @jinchihe I am working on updating secrets permissions in pipeline runner under Kubeflow manifest. If that works then there will not be any need to make the change here.

OK. Thanks.

PR #497, opened 2020-05-28
Add example notebook stitching together BentoML and Fairing

iancoffey:

I find the use cases around the combo of BentoML and KF pretty interesting, so I made this little experiment notebook to flesh out these ideas. Now that it is done, I thought my experiment notebook might make a good example for this project. I trained a Prophet model with covid19 time-series data for the experiment and then stitched the BentoML artifacts to Fairings Cluster builder.

I was also thinking about extending the end of the notebook to use KFServing instead of directly creating pods for the model apis, and contributing this instead to the kf-examples repo. 

Im very interested in folks thoughts in general, cheers.


iancoffey:
Im going to take my own suggestion and close and reopen after updating the deploy.

PR #501, opened 2020-06-09
Allow user specify existing image

jinchihe:

Sometimes, user wants to use an existing image to deployer (or build without Fairing), instead of rebuild image. The PR is going to allow user specify existing image.

And release new one in Pypi to include recent fixes. Thanks.


jinchihe:
/approve

Thanks @fenglixa

PR #505, opened 2020-06-17
Detach GPU from TFJob PS Spec

### LINKED ISSUE #504: Should not attach GPU resource for TFJob PS pod
/kind feature

[A clear and concise description of what you want to happen.]

If we set the GPU resource, we should not attach GPU resource for PS, just need to Chief and Worker node.

https://github.com/kubeflow/fairing/blob/804916bd71c32de45afa2a4ecba9607f4283e990/kubeflow/fairing/kubernetes/utils.py#L33

jinchihe:
1. Do not attach GPU resource for PS pod.
2. Update the Chief default value to 0, since sometimes, just need worker and PS.

Fixes #504

### DIFF
--- a/kubeflow/fairing/deployers/tfjob/tfjob.py
+++ b/kubeflow/fairing/deployers/tfjob/tfjob.py
@@ -1,4 +1,5 @@
 import logging
+import copy
 from kubernetes import client as k8s_client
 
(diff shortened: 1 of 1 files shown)


jinchihe:
/approve

Thanks @pugangxa

PR #515, opened 2020-06-24
Support configuring environment variables

### LINKED ISSUE #492: Fairing Serving job user can not set the custom environment vars
/kind feature

so far, for the Faring Serving Job , user can not set the specific custom environment vars like the custom command, is it possible let user to set the env parmameters like command in the preprocessor to build in the pod spec?

_[class BasePreProcessor(object)](https://github.com/kubeflow/fairing/blob/master/kubeflow/fairing/preprocessors/base.py):
    """Prepares a context that gets sent to the builder for the docker build and sets the entrypoint
    :param input_files: the source files to be processed
    :param executable: the file to execute using command (e.g. main.py)
    :param output_map: a dict of files to be added without preprocessing
    :param path_prefix: the prefix of the path where the files will be added in the container
    :param command: the command to pass to the builder_

jinchihe:
Support configuring environment variables
Fixes #492 

**Release note**:

```release-note

```

### DIFF
--- a/kubeflow/fairing/kubernetes/utils.py
+++ b/kubeflow/fairing/kubernetes/utils.py
@@ -66,3 +66,22 @@ def mounting_pvc(pvc_name, pvc_mount_path=constants.PVC_DEFAULT_MOUNT_PATH):
         else:
             pod_spec.volumes = [volume]
     return _mounting_pvc
+
+def add_env(env_vars):
+    """The function for pod_spec_mutators to add custom environment vars.
+
+    :param vars: dict of custom environment vars.
+    :returns: object: function for add environment vars to pods.
+
+    """
+    def _add_env(kube_manager, pod_spec, namespace): #pylint:disable=unused-argument
+        env_list = []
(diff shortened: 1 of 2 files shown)


IvySong2016:
the implement about how to append environment is ok, but there is no reference in the preprocessor for end user the specify the env pairs they want to set.

jinchihe:
> the implement about how to append environment is ok, but there is no reference in the preprocessor for end user the specify the env pairs they want to set.

Thanks. The pod spec is generated in builder. So that's better to set env via `pod_spec_mutators`. For example:

```
from kubeflow.fairing.deployers.job import Job
from kubeflow.fairing.kubernetes import utils

Job(namespace=xxx, pod_spec_mutators=[utils.add_env({'var1': 'value1', 'var2', 'value2', ...})])
```

fenglixa:
Thanks @jinchihe, 
/lgtm

xauthulei:
/lgtm
Thanks @jinchihe

PR #518, opened 2020-06-28
Update OWNERS

jinchihe:
From @jlewi

> A couple of things to call out for people who have been discussing forming WGs in various areas.
> 
> Please take a look at the roles defined for each WG and think about who you would want to assume those roles
> 
> Please think about ensuring your OWNERs files are up to date and adequately reflect who the OWNERs are for various sub projects. This will be critical to ensuring OWNERs don't get overlooked.
> 

Update OWINERS, remove some inactive owners, and add @xauthulei who has some contributors to review code, thanks.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,5 +1,3 @@
 approvers:
-  - abhi-g
   - jinchihe
-  - karthikv2k
-  - r2d4
+  - xauthulei


jinchihe:
@abhi-g @karthikv2k @r2d4 The PR will keep couple of days, please append if you want to keep in the OWNERS list. Thanks!

abhi-g:
@jinchihe please feel free to remove me from owners.

Thanks for your continued contributions to Fairing.

jlewi:
@jinchihe We should probably have a minimum of 2-3 approvers just so we have a backup for emergencies. If there isn't an active contributor who would make sense for this then the next best thing would be to add someone who can be trusted to only approve PRs in an emergency. 

Any suggestions?

jinchihe:
@jlewi Thanks! Agree with you, if so, I think we can add @xauthulei as approver. Please let me know if you think we should add others.

jlewi:
@jinchihe Sounds good to me

xauthulei:
/lgtm

Thanks

PR #524, opened 2020-07-08
Fix broken PredictionEndpoint

jer0nim0:
This PR fixes a bug in the constructor of PredictionEndpoint, which calls its base class constructor with wrong arguments.

The primary example referenced from the Kubeflow docs, xgboost-high-level-apis.ipynb, was broken before this fix.


jer0nim0:
@googlebot I signed it!

jinchihe:
Thanks for the contribution! @jer0nim0 

/lgtm
/approve

PR #530, opened 2020-08-13
Add --user option to pip install command to support non root user environment

Minkyu-Choi:
Currently, running cluster builder with non root user base image(e.g. gcr.io/kubeflow-images-public/tensorflow-2.1.0-notebook-cpu:1.0.0) results error when user tries to install additional python package using requirements.txt file. 

Pip install without --user option installs package under /usr/local/lib path. And non root user doesn't have right to write under /usr/local/lib and this lead to Permission Denied error for pip install.

With this modification user can install additional python package regardless of container's user.

Fixes #NONE

### DIFF
--- a/kubeflow/fairing/builders/dockerfile.py
+++ b/kubeflow/fairing/builders/dockerfile.py
@@ -30,7 +30,7 @@ def write_dockerfile(
     if install_reqs_before_copy:
         content_lines.append("COPY {}/requirements.txt {}".format(path_prefix, path_prefix))
     content_lines.append("RUN if [ -e requirements.txt ];" +
-                         "then pip install --no-cache -r requirements.txt; fi")
+                         "then pip install --user --no-cache -r requirements.txt; fi")
     content_lines.append(copy_context)
 
     if docker_command:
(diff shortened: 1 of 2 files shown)


xauthulei:
/ok-to-test

Thanks @Minkyu-Choi

Minkyu-Choi:
There's a PATH problem with root user.
Should we divide root and non root user case for write_dockerfile?

jinchihe:
I think that's OK to add the option. For root user, the packages should be also installed successfully.

/retest

jinchihe:
@Minkyu-Choi The CI test failed, seems we need to handle this

```
Step #2: INFO     kubeflow.fairing.builders.docker.docker:docker.py:103 Build output:   WARNING: The scripts jupyter-bundlerextension, ju
pyter-nbextension, jupyter-notebook and jupyter-serverextension are installed in '/root/.local/bin' which is not on PATH.
Step #2:   Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
Step #2:
```

Minkyu-Choi:
Sure, I agree. packages should be installed successfully for root user.
Pip install using --user option with root installs package executable to /root/.local/bin. I missed executable. 
Root user doesn't have /root/.local/bin in PATH variable. User option for root user makes no sense.

So, I'm considering use of whoami command to determine container's user and nested conditional statement for existence of requirements.txt and user type.

I'll going to open new PR when I'm ready. Thanks

PR #536, opened 2020-08-20
Add conditional statement to dockerfile considering user of base image

Minkyu-Choi:
As I wrote before(https://github.com/kubeflow/fairing/pull/530), if user of base image is non-root then pip install command should have --user option to avoid permission problem. Also, $HOME/.local/lib path should be included to PATH variable for package executable.

However, there's no easy way to get value of home path unless it is explicitly set by ENV directive. 
So, I added executable_path_prefix option to specify home path of non-root user.

If executable_path_prefix is given, ENV directive will be added to include $HOME/.local/lib.
Fairing user doesn't need to specify executable_path_prefix if $HOME/.local/lib path is already included in a base image.

Base image with root user will be processed same as before.

There can be a better way of considering user of base image.
Any ideas are welcome.

Fixes #None

None

**Release note**:

```release-note

```

### DIFF
--- a/kubeflow/fairing/builders/cluster/cluster.py
+++ b/kubeflow/fairing/builders/cluster/cluster.py
@@ -26,7 +26,8 @@ class ClusterBuilder(BaseBuilder):
                  pod_spec_mutators=None,
                  namespace=None,
                  dockerfile_path=None,
-                 cleanup=False):
+                 cleanup=False,
+                 executable_path_prefix=None):
         super().__init__(
             registry=registry,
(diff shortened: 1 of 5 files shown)


Minkyu-Choi:
@jinchihe @pugangxa Can this get reviewed?

jinchihe:
/ok-to-test

@Minkyu-Choi Thanks for the contribution.

jinchihe:
Looks great! @Minkyu-Choi 

/lgtm
/approve

Issue #452, opened 2020-02-10
No module named 'cloudpickle.cloudpickle_fast' when launching JOB


Luke035:
/kind bug

**What steps did you take and what happened:**
I launched a simple JOB made of a basic function (just print a statement)
```
from kubeflow import fairing
from kubeflow.fairing import TrainJob
from kubeflow.fairing.backends import KubeflowAWSBackend

AWS_ACCOUNT_ID = fairing.cloud.aws.guess_account_id()
AWS_REGION = 'us-east-1'
DOCKER_REGISTRY = '{}.dkr.ecr.{}.amazonaws.com'.format(AWS_ACCOUNT_ID, AWS_REGION)

def simple_function():
    print("Hello Fairing world")

job = TrainJob(simple_function,base_docker_image='AWS_ECR_REPO/python:3.8-slim', docker_registry=DOCKER_REGISTRY, backend=KubeflowAWSBackend())

job.submit()
```

After launching the submit command the container is built correctly and eventually pushed into the registry, but once launched  I get the following error

```
[W 200210 15:07:40 manager:255] Waiting for fairing-job-55nc4-rccmv to start...
[E 200210 15:07:41 manager:272] Failed to launch fairing-job-55nc4-rccmv, reason: Error, message: None
[W 200210 15:07:42 job:162] Cleaning up job fairing-job-55nc4...
Traceback (most recent call last):
  File "/app/function_shim.py", line 2, in <module>
    import cloudpickle
  File "/app/cloudpickle/__init__.py", line 9, in <module>
    from cloudpickle.cloudpickle_fast import CloudPickler, dumps, dump
ModuleNotFoundError: No module named 'cloudpickle.cloudpickle_fast'
```

**What did you expect to happen:**
Job correctly launched

**Anything else you would like to add:**
If I launched the container manually (`docker run ...`) I get the same error

**Environment:**

- Fairing version: 0.7.0
- Kubeflow version: (version number can be found at the bottom left corner of the Kubeflow dashboard): 0.7.0
- Minikube version: NA
- Kubernetes version: (use `kubectl version`): 1.14
- OS (e.g. from `/etc/os-release`): Amazon Linux

MinhManPham:
Hi @Luke035 , i got the same problem, are there any solutions for this ?. Thanks

PR #541, opened 2020-09-09
Fix for failing Training Job on py 3.8+

izapolsk:
I've encountered the issue similar to described in https://github.com/kubeflow/fairing/issues/452.
It occurs when user uses function preprocessor + cloudpicke version containing cloudpickle_fast.py + python 3.8+.
 
The root cause is that newer versions of cloudpickle contain cloudpickle_fast.py and that file isn't copied by preprocessor during image preparation.

So, this PR aims to append mentioned above file along with the rest of cloudpickle package.

By the way, what's the reason for copying that package into image ? Wouldn't it be better to copy egg there f.e. and install it on start up ?

### DIFF
--- a/kubeflow/fairing/preprocessors/function.py
+++ b/kubeflow/fairing/preprocessors/function.py
@@ -64,10 +64,11 @@ class FunctionPreProcessor(BasePreProcessor):
 
         # Make sure cloudpickle can be imported as a module
         cloudpickle_dir = os.path.dirname(cloudpickle.__file__)
-        self.output_map[os.path.join(cloudpickle_dir, '__init__.py')] = \
-            os.path.join(path_prefix, "cloudpickle", '__init__.py')
-        self.output_map[os.path.join(cloudpickle_dir, 'cloudpickle.py')] = \
-            os.path.join(path_prefix, "cloudpickle", 'cloudpickle.py')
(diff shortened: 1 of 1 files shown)


izapolsk:
@googlebot I signed it!

xauthulei:
Thanks @izapolsk

xauthulei:
Currently it report this issue:
```
Files with lint issues: /build/lib/kubeflow/fairing/preprocessors/function.py, /kubeflow/fairing/preprocessors/function.py
```
You could install the `Pylint` checking plugin in your python development env to catch the errors/warning here.

jinchihe:
@izapolsk This is caused by line-too-long,  easy to fix. Thanks

```
INFO|2020-09-10T01:45:19|util.py:72| ************* Module kubeflow.fairing.preprocessors.function
INFO|2020-09-10T01:45:19|util.py:72| C: 69, 0: Line too long (118/100) (line-too-long)
INFO|2020-09-10T01:45:19|util.py:72| 
INFO|2020-09-10T01:45:19|util.py:72| -----------------------------------
INFO|2020-09-10T01:45:19|util.py:72| Your code has been rated at 9.77/10
```

izapolsk:
sorry, my default setting is  =<120

PR #554, opened 2021-03-23
allow caching and overriding single_snapshot for kaniko

omrishiv:
In our environment, image builds take a long time because the base layer is large and our updates are small. We have to use the in-cluster builder and this allows us to snapshot and cache the intermediate layers so subsequent builds become faster. 

Fixes #

This PR should be backwards compatible with older versions.

**Release note**:

```release-note
Enables creation of a cluster builder with the ability to cache and not use single_snapshots to enable faster incremental builds in certain situations:
```
            cluster_builder = ClusterBuilder(registry=constants.DEFAULT_DOCKER_REGISTRY,
                                             base_image=self._base_image_name,
                                             preprocessor=preprocessor,
                                             image_name=image_name,
                                             dockerfile_path='Dockerfile',
                                             context_source=create_minio_context_source(),
                                             cache=True,
                                             single_snapshot=False)
```

```

### DIFF
--- a/kubeflow/fairing/builders/cluster/cluster.py
+++ b/kubeflow/fairing/builders/cluster/cluster.py
@@ -27,7 +27,9 @@ class ClusterBuilder(BaseBuilder):
                  namespace=None,
                  dockerfile_path=None,
                  cleanup=False,
-                 executable_path_prefix=None):
+                 executable_path_prefix=None,
+                 cache=False,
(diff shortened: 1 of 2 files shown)


xauthulei:
@OmriShiv ,  please accept the CLA for the contribution firstly, Thanks

omrishiv:
@xauthulei Thank you, working on it with the org

